Issue #24, opened 2026-01-05
Test failures on `torch-fourier-slice` related to MTF files in ttsim3d


mgiammar:
The test for `torch-fourier-slice` call the `ttsim3d` package to simulate a test volume from a downloaded example PDB file. However, running the pytests via

```bash
uv run pytest
```

yields the following errors on all slice extraction tests which depend on simulation, likely due to incorrect path resolution in the workspace for how ttsim3d was downloaded:

```
_______________________________________________ test_extract_central_slices_rfft_3d_symmetry_cases_large_volume[angle_pair7-cpu] ________________________________________________

device = 'cpu', angle_pair = ([45, 90, 0], [45, 90, 180])

    @pytest.mark.parametrize("device", DEVICES)
    @pytest.mark.parametrize("angle_pair", ORIENTATION_SYMMETRY_PAIRS)
    def test_extract_central_slices_rfft_3d_symmetry_cases_large_volume(
        device: str, angle_pair: tuple[list[float], list[float]]
    ):
        """Tests extract_central_slices_rfft_3d on symmetry-related orientations."""
>       volume = setup_slice_volume().to(device)
                 ^^^^^^^^^^^^^^^^^^^^

packages/primitives/torch-fourier-slice/tests/test_extract_central_slices_rfft_3d.py:153: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
packages/primitives/torch-fourier-slice/tests/test_extract_central_slices_rfft_3d.py:69: in setup_slice_volume
    return sim.run()
           ^^^^^^^^^
.venv/lib/python3.13/site-packages/ttsim3d/models.py:296: in run
    mtf_frequencies, mtf_amplitudes = self.simulator_config.mtf_tensors
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.13/site-packages/ttsim3d/models.py:131: in mtf_tensors
    frequencies, amplitudes = read_mtf(file_path=self.mtf_reference)
[...]

PR #25, opened 2026-01-05
Passing workspace unit tests

mgiammar:
Unit tests run via `uv run pytest` now pass for all included primitive packages. Note that `uv pip install -U ttsim3d` may need to be run separately from `uv sync` to properly update the ttsim3d package; this will be resolved with a new ttsim3d release.

This PR also does the following:

- Sets the fallback minimum version for all packages to `v0.5.0.` to ensure they exceed what is currently released on PyPI.
- Resolves an incompatible tensor float64 type on `mps` devices in the torch-grid-utils when unit tests are run on MacOS.
- Forces `uv` to use the workspace version of torch-fourier-filter for ttsim3d compatibility (see #24). This may need to be done for all other packages at some point.


alisterburt (approved):
incredible stuff! changing the fallback version is a much smarter solution than the tag pushing I was suggesting, great work

alisterburt:
I'm a little confused why the override-dependencies is required when the package is already specified as a dependency and as has its source set to workspace... weird but no biggie

PR #30, opened 2026-01-09
fix package discovery logic

alisterburt:
wip, need to see how it works in CI


alisterburt:
@mgiammar this is now working ish, most packages are green

few things that came up
- I think this will run for all packages right now on any change as any change will affect uv.lock triggering test of metapackage and therefore all packages -> I will eventually separate out metapackage tests from workspace package tests to fix this but not in this PR
- it seems once I push a tag for a single package the ability to build other packages without tags yet and drop to fallback version fails -> probably easiest fix is to push a tagged version of everything at once to avoid this
- tests which depends on ttsim3d install it fine but mtf file required for simulation isn't found, is this fixed in a later version or still an issue with ttsim3d?

mgiammar:
> tests which depends on ttsim3d install it fine but mtf file required for simulation isn't found, is this fixed in a later version or still an issue with ttsim3d?

@alisterburt Will need to push a new release of ttsim3d which relaxes a torch-fourier-filter version requirement. Currently only v0.3.0 is installable and falls back to a PyPI installation, but v0.4.0 fixes the MTF file not found error. Again, I think this is just a migration pain and should be better once packages start getting released from this teamtomo repo

alisterburt:
Cool! No rush, can always xfail the test here temporarily while we bring things up :-)

PR #31, opened 2026-01-10
Update dynamic versioning commands in build system

mgiammar:
After trying to get the `vcs` package versioning working locally with some package tags (e.g. `torch-grid-utils@v0.5.0`) _and_ after doing some research on what's possible with the current hatch build system, I've adjusted all sub-packages to use a fixed versioning system where `pyproject.toml` files are used to define package versions rather than from a git tag.

@alisterburt This is not optimal for package ease-of-release since a developer would have to 1) bump the package versions manually, then 2) use some tagging mechanism to trigger a new release. Some CI mechanism could verify that a tagged version matches what gets tagged.

## Todo before merging

 - [x] Verify dynamic versioning for `teamtomo` metarepo still works
 - [ ] Test out CI release mechanism with new versioning system
 - [x] Update development notes accordingly

### DIFF
--- a/notes/migrate-existing-repo.md
+++ b/notes/migrate-existing-repo.md
@@ -341,6 +341,9 @@ teamtomo/                            # Monorepo root
 
 ### Version Configuration
 
+Teamtomo uses a dynamic versioning process through `hatch-vcs` to detect git tags with a package's specific version; these git tags follow a scheme of `<package-name>@v<x.y.z>`.
+Ensure the `[tool.hatch.version]` and `[tool.hatch.version.raw-options]` tables are updated accordingly:
+
 **Before (standalone):**
 
 ```toml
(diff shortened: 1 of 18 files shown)


alisterburt:
Thanks for this - I agree it's suboptimal, before we push ahead with this could we check what happens if we tag all packages? I know that individual packages picked up the version fine from vcs so wonder if it's just a weird failure mode we can avoid by tagging all at once...

If that ends up not working then of course we can go ahead with this :-)

mgiammar:
```
[stderr]
      /Users/mgiammar/.cache/uv/builds-v0/.tmpnSDX5G/lib/python3.12/site-packages/setuptools_scm/version.py:108: UserWarning: tag 'torch-grid-utils@v0.5.0' no
      version found
        warnings.warn(f"tag {tag!r} no version found")
...
```

Unfortunately same vcs error with all packages tagged, at least on my laptop.

alisterburt:
Agh - one last Hail Mary before we give up? https://github.com/ofek/hatch-vcs/issues/71

mgiammar:
@alisterburt This setup now works on my system by explicitly defining the git command to use for inspecting tags. Odd behaviors, but hopefully will work with CI

alisterburt (approved):
🥳🥳🥳 

really glad this worked out!

alisterburt:
let's merge this then try to push 0.5.0 of grid utils as a first attempt! 

Great if you can do that this morning, otherwise I'll get to it a bit later today :-)

PR #54, opened 2026-03-17
Change to a local script for tagging and release of packages

mgiammar:
Instead of relying on a GitHub actions + app to tag releases for each of the releases (is fragile, and have not been able to get one workflow to trigger another workflow...), this PR changes to a local script for creating and pushing all the tags.

Pinging @alisterburt @McHaillet @jdickerson95 in case there are any opinions on this change. Maybe there is a future where everything is handled by an actions workflow, but this seems like the most practical near-term solution.


McHaillet (commented):
Makes a lot of sense to me. Only thing when reading through the release instructions file: for the single release it should only be allowed to increment the patch version. Is there a specific restriction on this? It could check against the latest released version and only deploy if the patch version is the only things that increases. (sorry for my ignorance if this is already implemented)

mgiammar:
For now, I think it's easiest to have the core devs do the version releases and have a more "manual" process of choosing a new version. Could be automated and checked in the future, but I think less is more towards getting the monorepo up and functioning right now.

PR #66, opened 2026-04-13
Migrate changes of torch-fourier-slice to monoreop

mgiammar:


mgiammar:
@McHaillet I had to increase the tolerance on the last three tiltxcorr unit test to get those passing. They were just above the previously set `0.5` threshold. Some interaction with the new default padding parameters in torch-fourier-slice. Don't think there's any real issue here.

PR #77, opened 2026-04-14
Adding 3D convenience layer

sjrothfuss:
Pulling over [this PR](https://github.com/teamtomo/torch-transform-image/pull/15) from the independent repo to the monorepo.

### DIFF
--- a/packages/primitives/torch-transform-image/src/torch_transform_image/__init__.py
+++ b/packages/primitives/torch-transform-image/src/torch_transform_image/__init__.py
@@ -9,12 +9,22 @@ except PackageNotFoundError:
 __author__ = "Alister Burt"
 __email__ = "alisterburt@gmail.com"
 
-from torch_transform_image.transforms_2d import affine_transform_image_2d, rotate_then_shift_image_2d, shift_then_rotate_image_2d
-from torch_transform_image.transforms_3d import affine_transform_image_3d
+from torch_transform_image.transforms_2d import (
+    affine_transform_image_2d,
+    rotate_then_shift_image_2d,
+    shift_then_rotate_image_2d,
+)
+from torch_transform_image.transforms_3d import (
+    affine_transform_image_3d,
+    rotate_then_shift_image_3d,
+    shift_then_rotate_image_3d,
+)
 
 __all__ = [
-    'affine_transform_image_2d',
-    'rotate_then_shift_image_2d',
-    'shift_then_rotate_image_2d',
-    'affine_transform_image_3d',
+    "affine_transform_image_2d",
+    "rotate_then_shift_image_2d",
+    "shift_then_rotate_image_2d",
+    "affine_transform_image_3d",
+    "rotate_then_shift_image_3d",
+    "shift_then_rotate_image_3d",
 ]
--- a/packages/primitives/torch-transform-image/src/torch_transform_image/transforms_2d.py
+++ b/packages/primitives/torch-transform-image/src/torch_transform_image/transforms_2d.py
@@ -164,13 +164,14 @@ def _build_rotate_shift_matrix_2d(
         e = f"2 shifts are required but {num_shifts} were supplied: {shift}"
         raise ValueError(e)
 
-    rotation_matrix = R([rotate])
-    # Because shift is applied to the coordinate grid, it must be
-    # negated to produce a positive (up/right) shift on the image.
-    translation_matrix = T(-torch.as_tensor(shift, dtype=torch.float32))
+    rotation_matrix = R([rotate], yx=True)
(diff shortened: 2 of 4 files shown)


jdickerson95 (commented):
Sorry for the delay in getting to this @sjrothfuss 
I generally like this to match the 2D file.

One nitpick I have is that "shifts_zyx" should probably be renamed "shift_zyx", unless multiple shifts are supported.

My other thought is for both the 2D and 3D. Could there be a private helper function to remove duplicate code? Something like this for the 3D:

```
def _rotate_shift_matrix_3d(
    rotate_zyx, shift_zyx, image_center, *, rotate_first: bool
) -> torch.Tensor:

    Validation things common to both here...

    Rmat = Rx(rotate_zyx[2], zyx=True) @ Ry(rotate_zyx[1], zyx=True) @ Rz(rotate_zyx[0], zyx=True)
    Tshift = T(shift_zyx)
    if rotate_first:
        inner = Tshift @ Rmat
    else:
        inner = Rmat @ Tshift
    return torch.inverse(T(image_center) @ inner @ T(-image_center))
```

And then the two public functions you added become thin wrappers to pass the boolean into this function.

sjrothfuss:
Thanks, Josh! I like both of those changes. Having a hidden wrapper ameliorates both @alisterburt and I's concerns from https://github.com/teamtomo/torch-transform-image/pull/10#issuecomment-2835606254; there'd be two user-facing functions but still only one code block to maintain. I'm traveling this week but barring any objections I'll implement those changes.

sjrothfuss:
Happy Friday @jdickerson95. I added private helpers to both 2d and 3d and updated the 2d functions to be a little more similar to 3d. Happy to hear any other feedback!

mgiammar (approved):
Looks good to me @sjrothfuss. @jdickerson95 @McHaillet any comments before merging?

McHaillet:
Hey! Thanks for the continued effort on this :)

The helper functions look inconsistent with each other. This is the helper for 2d:

```python
def _build_rotate_shift_matrix_2d(
    ....
    rotation_matrix = R([rotate])
    # Because shift is applied to the coordinate grid, it must be
    # negated to produce a positive (up/right) shift on the image.
    translation_matrix = T(-torch.as_tensor(shift, dtype=torch.float32))

    if rotate_first:
        inner_matrix = rotation_matrix @ translation_matrix
    else:
        inner_matrix = translation_matrix @ rotation_matrix
    return T(center_tensor) @ inner_matrix @ T(-center_tensor)
```

and for 3d:

```python
def _build_rotate_shift_matrix_3d(
    ...
    rotation_matrix = (
            Rx(rotate_zyx[2], zyx=True)
            @ Ry(rotate_zyx[1], zyx=True)
            @ Rz(rotate_zyx[0], zyx=True)
        )
    translation_matrix = T(shift_zyx)

    if rotate_first:
        inner_matrix = translation_matrix @ rotation_matrix
    else:
        inner_matrix = rotation_matrix @ translation_matrix
    matrix = T(image_center) @ inner_matrix @ T(-image_center)
    # Matrix is inverted because it is applied to the coordinate grid,
    # not the image directly.
    return torch.inverse(matrix)
```

But going through the tests they both behave correctly.

I think if you call the 2D rotation matrix with yx=True, remove the shift negation, and instead invert the matrix at the end they are still consistent but also look the same :+1: 

Also noticed in the 3D versions the variable is shift_zyx, and in 2d its just shift, so I changed it to shift_yx.

McHaillet:
~~I also don't understand how [this](https://github.com/sjrothfuss/teamtomo/blob/8853b3c6ee8a7c3ff8c37f3f214cf4bea9423049/packages/primitives/torch-transform-image/tests/test_torch_image_transform.py#L157) can simultaneously be true in the test :upside_down_face: :~~

```python
    assert image[14, 14, 26] == 0
    assert torch.allclose(result[14, 14, 26], torch.tensor(1.0), atol=1e-6)
```

edit: Nevermind, i'm an idiot, didn't read correctly that one was source and the other the result

McHaillet:
Transform image tests are working now, but the test suite failed due to a connection timeout in predict_fiducial_mask. It downloads data from Zenodo and probably hit a limit due to me triggering the test suite multiple times in a row.

To me this is good to go now. ~~Only thing is the slightly strange double assert I mentioned above.~~

McHaillet:
Now it runs :)

sjrothfuss:
All good changes @McHaillet. Remaining the 2d argument to `shift_yx` and getting consistent operations were on my todo list for eventually. Thanks for nabbing those.

McHaillet:
No problem! Merging it in :)

PR #78, opened 2026-04-15
Simplify motion estimation backend by breaking apart into helper classes/functions

mgiammar:
Have gone through and refactored some of the backend to pull out common functions and group together data into dataclasses. Makes the method easier to reason about IMO and could be nice for downstream integration with other packages. 

Currently implemented tests pass; have not tested a full motion correction pipeline. Also would like to write some targeted unit tests for new helper methods/classes.

Pinging @jdickerson95


jdickerson95:
I haven't tested it properly yet to identify all the breaking changes.
One question I have is, should the patch iterator go in torch-grid-utils? It seems like it would be generally useful and fit in the existing patch grid there.

mgiammar:
@jdickerson95 Will test these changes out tomorrow on some movie motion correction. Should give a surface-level idea of what needs updated.

I do think a patch iterator like function could go in `torch-grid-utils`. One distinction this PR and that discussion brings up is where helper data classes should live (`PatchIterator` is a class here), namely if they're appropriate for the primitives packages. I do like how the primitives are simple and structured as tensor + scalar only, no complex class relations.

The `PatchIterator` could change to a thin wrapper around a `torch-grid-utils` function which could be nice. Again, just thoughts on infrastructure design.

packages/algorithms/torch-motion-correction/src/torch_motion_correction/deformation_field_utils.py
```diff
@@ -1,162 +1,71 @@
...
+    | CubicCatmullRomGrid3d
+    | CubicBSplineGrid3d,
+    t: float,
+    grid_shape: tuple[int, int],
     grid_type: str = "catmull_rom",
 ) -> torch.Tensor:
-    """Evaluate a grid of shifts at a specific timepoint from deformation field data.
+    """Evaluate a dense shift grid at a single normalized timepoint."""
+    if isinstance(deformation_field, (CubicCatmullRomGrid3d, CubicBSplineGrid3d)):
+        import einops
```
jdickerson95: imports should not be in a function

packages/algorithms/torch-motion-correction/src/torch_motion_correction/correct_motion.py
```diff
@@ -9,18 +9,14 @@
...
-    evaluate_deformation_field,
-    evaluate_deformation_field_at_t,
-)
+from torch_motion_correction.types import DeformationField
 
 
 def correct_motion(
     image: torch.Tensor,  # (t, h, w)
-    deformation_grid: torch.Tensor,  # (yx, t, h, w)
+    deformation_field: DeformationField,
```
jdickerson95, line 16: The class may help organizationally for dealing with inputs, but it also adds an extra layer of complexity.
Being able to use this function through a simple tensor was simpler imo.

All other functions in this file still use a tensor, they should be consistent either way as well.

packages/algorithms/torch-motion-correction/src/torch_motion_correction/data_io.py
mgiammar, line 1: This file could be deleted and move all functionality into the DeformationField class

packages/algorithms/torch-motion-correction/src/torch_motion_correction/deformation_field_utils.py
mgiammar, line 1: Again, file could be deleted to instead rely on `DeformationField` class or associated helper functions

packages/algorithms/torch-motion-correction/src/torch_motion_correction/estimate_motion_optimizer.py
```diff
@@ -206,237 +130,291 @@ def estimate_local_motion(
...
+        b_factor_envelope=b_factor_envelope,
+        bandpass=bandpass_filter,
+        base_deformation_field=deformation_field,
+        new_deformation_field=new_deformation_field,
+        pixel_spacing=pixel_spacing,
+        ph=ph,
+        pw=pw,
+        loss_type=loss_type,
+        t=t,
+    )
```
mgiammar: On second thought, the `functools.partial` wrapper obscures what's going on under the hood. Should revert this in favor of explicit kwargs

packages/algorithms/torch-motion-correction/src/torch_motion_correction/estimate_motion_optimizer.py
```diff
@@ -206,237 +130,291 @@ def estimate_local_motion(
...
+    motion_optimizer.zero_grad()
+    return total_loss / n_batches if n_batches > 0 else 0.0
+
+
+def _initialize_deformation_grids(
+    resolution: tuple[int, int, int],
+    initial_deformation_field: DeformationField | None,
+    grid_type: str,
+    device: torch.device,
+) -> tuple[
```
mgiammar: Function seems like it should live somewhere in in the DeformationField logic rather than the optimizer logic

packages/algorithms/torch-motion-correction/src/torch_motion_correction/estimate_motion_xc.py
```diff
@@ -382,33 +339,41 @@ def estimate_motion_cross_correlation_patches(
...
     print(
         f"Estimated deformation field range: "
-        f"y=[{deformation_field[0].min():.1f}, {deformation_field[0].max():.1f}], "
-        f"x=[{deformation_field[1].min():.1f}, {deformation_field[1].max():.1f}]"
+        f"y=[{def_min_y:.1f}, {def_max_y:.1f}], "
+        f"x=[{def_min_x:.1f}, {def_max_x:.1f}]"
     )
 
-    print(f"Estimated deformation field shape: {deformation_field.shape}")
+    print(f"Estimated deformation field shape: {deformation_field_tensor.shape}")
```
mgiammar: Should wrap a verbose option around print statements

packages/algorithms/torch-motion-correction/src/torch_motion_correction/correct_motion.py
mgiammar: My reasoning behind a helper class is that the tensor data is also coupled to a grid type. Plus sampling logic can be placed in the helper class itself.

Can see what it looks like if properly consistent across the file. Can also revert changes too

mgiammar: Following our discussion this morning, moved everything over to the DeformationField class.

jdickerson95 (approved):

PR #86, opened 2026-05-12
fix(torch-fourier-rescale): preserve mean when source and target have different parity

amineuron:
## Summary

`fourier_rescale_2d` / `fourier_rescale_3d` with `preserve_mean=True` (the default) silently collapse the output mean to ~0 instead of preserving the input mean when the source and target spatial dimensions have different parity (e.g. even source, odd target).

## Reproducer

```python
import torch
from torch_fourier_rescale import fourier_rescale_2d

image = torch.full((16, 16), 5.0)
out, _ = fourier_rescale_2d(image, target_shape=(7, 7), preserve_mean=True)
print(out.mean().item())  # actual: ~0.0    expected: ~5.0
```

I hit this in the wild in a cryo-EM tilt-stack prebin: 4096² at 1.49 Å rescaled to 10 Å (target 609², odd) — the resulting "image" had mean ≈ -3e-4 instead of the input mean of ~0.4, which is dramatically broken.

## Root cause

In `fourier_rescale_dimension` (full-FFT path), the crop math was:

```python
total_crop = current_dft_size - target_dim_length
crop_start = total_crop // 2  # ← placement is wrong when parity differs
```

That puts the DC bin at `source // 2 - crop_start` of the cropped array. The caller then applies `torch.fft.ifftshift(..., dim=-2)` which expects the DC at index `target // 2`. The two positions coincide when source and target share parity, but differ by 1 when they don't — and a 1-bin offset in the unshifted DFT sends the DC to a high-frequency bin, so the reconstructed image has near-zero mean instead of the input mean.

The padding path had the matching bug, papered over with an ad-hoc `if source_dim_length % 2 == 1: pad_end -= 1` workaround that gives the wrong total padding when target is odd.

## Fix

Place the DC at `target // 2` of the new array in both crop and pad paths:

```python
crop_start = current_dft_size // 2 - target_dim_length // 2
pad_start  = target_dim_length // 2 - current_dft_size // 2
```

Symmetric, parity-agnostic, and trivially correct against the ifftshift the caller applies right after. Same one-liner for both directions; the ad-hoc odd-source workaround in the pad path can go away.

## Test plan
- [x] Added `test_fourier_rescale_{2,3}d_preserves_mean_for_all_parities` covering every even/odd source × target combo plus the 4096→609 cryo-EM case
- [x] Verified the new tests fail on `main` (3 fail) and pass with the fix (10/10 pass)
- [x] All 21 pre-existing tests in `test_torch_fourier_rescale.py` still pass (31/31 total)
- [x] No API change

### DIFF
--- a/packages/primitives/torch-fourier-rescale/src/torch_fourier_rescale/utils.py
+++ b/packages/primitives/torch-fourier-rescale/src/torch_fourier_rescale/utils.py
@@ -88,13 +88,20 @@ def fourier_rescale_dimension(
             pad_spec = [0, 0] * (dft.ndim - dim - 1) + [0, pad_size] + [0, 0] * dim
             return F.pad(dft, pad_spec, mode="constant", value=0)
     else:
-        # For regular FFT dimensions (full frequency spectrum)
+        # For regular FFT dimensions (full frequency spectrum), the DC bin of
+        # an fftshift'd array of size N is at index `N // 2` and the upstream
+        # `torch.fft.ifftshift` expects it there too. The crop / pad must
+        # therefore leave the DC at position `target // 2` of the new array,
+        # not at `(source - target) // 2` from the start — those differ when
+        # source and target have different parity (e.g. even source, odd
+        # target). Using `total_crop // 2` (or `total_pad // 2`) silently
+        # placed the DC at the wrong index in those cases, sending it to a
+        # high-frequency bin after ifftshift and zeroing the mean of the
+        # reconstructed image.
         if target_dim_length < source_dim_length:
-            # Crop: Remove frequencies symmetrically from both ends
-            total_crop = current_dft_size - target_dim_length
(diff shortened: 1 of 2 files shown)


mgiammar:
Hey @amineuron thanks for finding and opening a PR for this bug! I went through and added test cases for rfft cases just to make sure there is no interaction with the DC component in those cases. Otherwise the code looks good to me.

Tagging @teamtomo/core-devs for visibility.

mgiammar (approved):

alisterburt:
Thanks! Looks great :-)

amineuron:
Thank you both! And thanks @mgiammar for adding the rfft coverage.
For context, I hit this while running a MissAlignment/Warp workflow, where parity-mismatch rescaling during stack preparation/downsampling can corrupt the intensity before reconstruction.

PR #88, opened 2026-06-08
Feat/volume align

rsanchezgarc:
Algorithm to fit maps or atomic models into another map.

It uses first a brute force search, and then it refines the solutions using gradient descent


packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/__init__.py
```diff
@@ -0,0 +1,303 @@
...
+            resolved_gradient = resolved_gradient.model_copy(
+                update={"pixel_size_angstroms": pixel_size_angstroms}
+            )
+
+        devices = resolved_gradient.devices
+        if devices is None:
+            devices = [str(reference.device)]
+
+        n_start = len(candidates)
+        from concurrent.futures import ThreadPoolExecutor
```
McHaillet: Move import to top of file

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/__init__.py
```diff
@@ -0,0 +1,303 @@
...
+        Alignment result from :func:`align_volumes` or :func:`exhaustive_search`.
+    interpolation : {"trilinear", "nearest"}
+        Interpolation mode.
+
+    Returns
+    -------
+    aligned : torch.Tensor
+        ``(d, h, w)`` mobile volume transformed to match the reference frame.
+    """
+    from torch_transform_image import affine_transform_image_3d
```
McHaillet: Also move to top? Or is there good reason to have it here?

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/__init__.py
```diff
@@ -0,0 +1,303 @@
...
+    M_rot[:3, :3] = R_3x3
+    M_rot[:3, 3] = centre - R_3x3 @ centre
+
+    # Translation 4x4: [I, -t; 0, 1]
+    T_t = torch.eye(4, dtype=torch.float32, device=device)
+    T_t[:3, 3] = -t
+
+    # Combined: rotate around centre, then shift by t
+    M_combined = M_rot @ T_t
+    return affine_transform_image_3d(
```
McHaillet, line 206: Just double checking: this function produces the correct alignment?
Hard for me to judge by just reading the matrices.

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/__init__.py
```diff
@@ -0,0 +1,303 @@
...
+from ._result import AlignmentResult
+from ._simulate import DEFAULT_SIMULATOR, DensitySimulator
+
+# Sentinel for "use default gradient config" (avoids mutable default argument)
+_GRADIENT_UNSET: object = object()
+
+
+def align_volumes(
+    reference: torch.Tensor,
+    mobile: torch.Tensor,
```
McHaillet: `mobile` read a bit strange to me at first glance, but as a pair `reference`/`mobile` seems fairly clear

McHaillet (commented):
Super nice to see this added 👍 Didn't have time to go through the full thing yet, so just had a look at the API. 

You now have these two entries:

`align_volumes(reference, mobile, ...)`
`align_map_to_pdb(density_map, pdb_path, ...)`

The two entry signatures could be more intuitive as something like this (also considering the package name):

`fit_map_in_map(mobile_map, reference_map, ...)`
`fit_map_in_pdb(mobile_map, reference_pdb, ...)`

And the inverse could also be nice to have `fit_pdb_in_map()`? It could just be a wrapper of course that inverses the transformation.

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/__init__.py
rsanchezgarc: I tested it with a few cases. Not claiming it is bug-free, but seemed to work when I checked the results in ChimeraX

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/__init__.py
rsanchezgarc: It is there for some legacy reasons, not relevant any longer. Will fix it

rsanchezgarc:
@McHaillet, I have just implemented the changes you requested and have implemented some additional tests. Let me know what you think

McHaillet:
Thanks for the updates! I'm at a conference this week, so probably won't have time to review, but I will definitely take a look after

mgiammar:
Hey @rsanchezgarc very cool PR! I also have not had time to fully go though the code yet, but one thing I did notice is there's some overlap of functionality from the `_simulate.py` file, the `torch-electrostatic-potential` package, and the `ttsim3d` package (external to the monorepo). This new algorithm might be a good driver to consolidate some of this functionality into the monorepo (not saying you have to do this at all!)

Another thought I had is that the programatic nature of the `_cli` module is straying slightly outside the objectives of the monorepo/teamtomo for core infrastructure. One approach to supplying a CLI would be have an algorithm `torch-fit-into-map` in the monorepo which the following API objectives:

> Given well formed (pre-processed) volumes or PDB files, align a mobile structure into a reference via user-functions `fit_{pdb,map}_into_{pdb,map}`. These user-functions compose lower-level teamtomo primitives to return some alignment result. Algorithms package also supplies some minimal pre-processing API for PDB/MRC files necessary for fitting.

Then, a separate CLI package could wrap this core functionality to make an end-user program which exists separately from the alignment algorithms themselves.

Still a really cool package idea and happy to chat more about design decisions and hear others' feedback!

jdickerson95:
Hi @rsanchezgarc , I love this PR!

I agree with @mgiammar that there is a lot of redundancy here with torch-calculate-electrostatic-potential and ttsim3d. I would also add torch-structure-manipulation into that ( https://github.com/teamtomo/torch-structure-manipulation/tree/jd_refactor). It's definitely a driver for us to consolidate things. 

Another thing I'm not sure about is providing the pdb's as strings and using gemmi to read them. I think I prefer them to be a dataframe, meaning a user could provide them from mmdf or gemmi.

rsanchezgarc:
Hi @jdickerson95 and @mgiammar . Thanks for the review. 
What do you think the roadmap should be?
I don't have strong preferences, but, judging from your comments, it seems that having torch-electrostatic-potential and ttsim3d consolidated, looks like the first think that should be done. Do you agree?

With respect to the CLI, I am happy to remove it and have a wrapper package somewhere else. Would you prefer to have a general teamtomo-cli repository where several algos can be called?

Same with respect to  torch-structure-manipulation. I am happy to use it.

McHaillet:
Hey @rsanchezgarc, sorry for taking some time to get back to this. We actually had a quick chat about this in the teamtomo meeting (for your awareness they now take place every first Wednesday of the month, always welcome to join!). 

The conclusion was: let's merge this sooner than later and iterate on certain aspects post-merge. The main things to update, revolve around duplicate functionality in the repo, and I listed what I think should be done before/after merging this PR:
* use torch-electrostatic-potential for atom to density calculation [after merge]
* use torch-structure-manipulation for rotating atoms sets [after merge]
* CLI upstream [before merge]
* passing pdb as dataframe [before merge]

Having the CLI upstream, allows you to move mrc reading / pdb reading to the CLI wrapper. Then here the functions just get tensors for density maps and dataframes for atomic coordinates, which minimizes the amount of code and external dependencies. 

I am not knowledgeable about whether there is one default dataframe format for atomic coordinates. I hope gemmi and mmdf are consistent with each other. Maybe @jdickerson95 can comment on that.

rsanchezgarc:
Hi @McHaillet and @jdickerson95 . I have just finished implementing the pre-merge requests. Can you check if everything is OK now? Btw, The CLI is now at https://github.com/rsanchezgarc/torch-fit-in-map-cli

McHaillet:
Thanks for updating, that already reduces the complexity a bit. For my part good-to-go.

McHaillet (approved):

McHaillet:
I'll merge it in. And I will add an issue to track the post-merge tasks.

PR #92, opened 2026-06-29
feat(torch-fourier-filter): add optional frequency_grid_px

joelyeois:
Allow callers to pass a pre-computed frequency grid in Å⁻¹ to b_envelope, dose_envelope, Cs_envelope, and Cc_envelope via the new frequency_grid_px keyword argument.

Motivation: these functions may be often called repeatedly from a stateful class that already holds a cached frequency grid (e.g. registered as a buffer on a nn.Module). Previously, each call would unconditionally allocate a new grid tensor even though the shape and pixel size never change. Passing frequency_grid_px skips that allocation entirely.

Backwards compatibility: fully preserved. image_shape and pixel_size work exactly as before when frequency_grid_px is not provided. A ValueError is raised only if neither is supplied.

### DIFF
--- a/packages/primitives/torch-fourier-filter/src/torch_fourier_filter/envelopes.py
+++ b/packages/primitives/torch-fourier-filter/src/torch_fourier_filter/envelopes.py
@@ -5,14 +5,35 @@ from torch_ctf import calculate_relativistic_electron_wavelength
 from torch_grid_utils.fftfreq_grid import fftfreq_grid
 
 
+def _b_envelope(frequency_grid_px: torch.Tensor, B: float) -> torch.Tensor:
+    """
+    Compute a B-factor envelope from a pre-computed frequency grid.
+
+    Parameters
+    ----------
+    frequency_grid_px: torch.Tensor
+        Frequency grid in Å⁻¹, e.g.
+        ``fftfreq_grid(image_shape, norm=True) / pixel_size``.
+    B: float
+        The B-factor value.
+        Suggested value is 5 A^2 / e-/A^2
+
+    Returns
+    -------
+    torch.Tensor
+        B-factor envelope
+    """
+    divisor = 4  # this is 4 for amplitude, 2 for intensity
+    return torch.exp(-(B * frequency_grid_px**2) / divisor)
+
+
 def b_envelope(
     B: float,
-    image_shape: tuple[int, int] | tuple[int, int, int] | None = None,
-    pixel_size: float | None = None,
+    image_shape: tuple[int, int] | tuple[int, int, int],
+    pixel_size: float,
     rfft: bool = True,
     fftshift: bool = False,
     device: torch.device = None,
-    frequency_grid_px: torch.Tensor | None = None,
 ) -> torch.Tensor:
     """
     Create a B-factor envelope for a Fourier transform.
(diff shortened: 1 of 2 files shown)


alisterburt:
Hi @joelyeois ! Great to have you here 🙂

This is definitely a problem but I don't love this solution, the APIs become messy (which params correspond to which path for construction? Passing a grid makes most other params irrelevant)

I would suggest instead adding a lower level layer which computes these various functions on the frequency grid directly, then having these higher level functions construct the grid and pass it to those lower level funcs 

Does that make sense? 

(side note, it's quite nice not feeling nearly as bad about asking for code changes now that the cost of producing code is so cheap!)

alisterburt:
Oh, it's early AM here, if I've said something stupid please tell me

alisterburt:
To be clear: the shape of that lower level layer should be one func per op

joelyeois:
Hi @alisterburt, thanks for the feedback! Apologies I do realize that my initial suggestion was quite messy. Just want to make sure I understand the suggested approach before reworking this.

Are you suggesting something like splitting each function into a lower-level `_b_envelope(frequency_grid_px, B)` that just does the math, and keeping the existing `b_envelope(image_shape, pixel_size, ...)` as a wrapper that constructs the grid and delegates to it? Callers with a cached grid would then skip the high-level function entirely and call `_b_envelope` directly.

Something like:
```
def _b_envelope(frequency_grid_px: torch.Tensor, B: float) -> torch.Tensor:
    divisor = 4
    return torch.exp(-(B * frequency_grid_px**2) / divisor)

def b_envelope(B, image_shape, pixel_size, rfft=True, fftshift=False, device=None):
    frequency_grid_px = fftfreq_grid(...) / pixel_size
    return _b_envelope(frequency_grid_px, B)
```

alisterburt:
@joelyeois exactly :-) 

restated/refined:
- math utils for calculating each property should be rank polymorphic as far as possible (i.e. not unnecessarily restricted to specific shapes)
- envelope funcs should call into those math utils
- for your use case you would directly use the math util on your precomputed grid

I'm not sure I understand the `frequency_grid_px` nomenclature, for me the units are either cycles/px or cycles/ang, but it looks like frequency_grid_pix here is in cycles/ang?

edit: I know it wasn't you who implemented the existing nomenclature, was it @jdickerson95 who added these? maybe he can jump in

mgiammar:
Quickly popping in here (and also relevant to @jdickerson95) frequency grids may not always be scaled uniformly by the pixel spacing. Anisotropic magnification can be modeled by stretching/contracting the frequency grid along some axes, but the underlying mathematics would remain the same. Hence separation of frequency grid definitions from math implementations.

jdickerson95:
@alisterburt, in my head, it was frequency_grid_px was frequency_grid (cycles/px) divided by pixel size. In retrospect, that is a little confusing.

alisterburt:
ahhh

joelyeois:
So shall I just keep to the same naming and recommit the new changes?

jdickerson95:
I don't have a strong opinion about the naming. I would keep it the same, but I'm happy for it to change if @alisterburt or @mgiammar have a preference.

mgiammar:
No strong preference on frequency grid naming conventions from my end.

packages/primitives/torch-fourier-filter/src/torch_fourier_filter/envelopes.py
```diff
@@ -5,6 +5,28 @@
...
+    Compute a B-factor envelope from a pre-computed frequency grid.
+
+    Parameters
+    ----------
+    frequency_grid_px: torch.Tensor
+        Frequency grid in Å⁻¹, e.g.
+        ``fftfreq_grid(image_shape, norm=True) / pixel_size``.
+    B: float
+        The B-factor value.
+        Suggested value is 5 A^2 / e-/A^2
```
mgiammar: Remove suggested value since could be arbitrary

packages/primitives/torch-fourier-filter/src/torch_fourier_filter/envelopes.py
```diff
@@ -103,12 +156,53 @@ def dose_envelope(
...
+    defocus *= 1e4  # microns -> angstroms
+
+    return torch.exp(
+        -(((torch.pi * (alpha / 1000)) / _lambda) ** 2)
+        * (
+            Cs * _lambda**3 * frequency_grid_px**3
+            + _lambda * (defocus) * frequency_grid_px
+        )
+        ** 2
+    )
```
mgiammar: Can this be written out in a multi-line statement with temporary variables? Otherwise somewhat difficult to quickly see what computation is being done.

mgiammar (changes requested):
Only very minor formatting requests, otherwise code looks good to me.

joelyeois:
Fixed both in d80a952 — dropped the suggested B-factor value and broke the Cs envelope math into named terms (`aperture_term`, `spherical_term`, etc.) so it's easier to read. Both were already in the code before this PR, just carried over in the split — hope this cleans it up!

PR #95, opened 2026-07-07
Add torch-scattering: multislice electron scattering simulation

joelyeois:
## Summary

Adds `torch-scattering`, a PyTorch package for simulating 2D electron exit waves from a 3D scattering potential, as a new `primitives/` package.

Four propagation modes are provided, trading physical accuracy for speed:

- `multislice()` — full multislice propagation (Kirkland), the most accurate mode
- `rytov()` — Rytov approximation
- `firstborn()` — first Born approximation
- `projection()` — projection approximation, the fastest and least accurate mode

All four share the same signature, support arbitrary batch dimensions, and accept an `n_slices` argument to coarsen the potential before propagating. Lower-level primitives (`fresnel_propagator`, `transmission_function`, `multislice_step`, `chunk_slices`, `interaction_parameter`) are also exposed for building custom propagation schemes.

Depends on `torch-grid-utils` and `torch-ctf` (both existing workspace members).

## Test plan

- [x] `uv sync --all-extras --all-packages` resolves the workspace with `torch-scattering` included
- [x] `uv run pytest packages/primitives/torch-scattering/tests/` — 44 passed
- [x] `ruff check` / `mypy` clean on the package source
- [x] `uv build packages/primitives/torch-scattering/` produces sdist/wheel
- [x] Import works both directly (`import torch_scattering`) and via `teamtomo.primitives.torch_scattering`


mgiammar:
Hey @joelyeois thanks again for taking the time to put this together! From a quick look, the API is nice and I like the similarities across the multiple methods. Only minor comment is consistency of parameter names across teamtomo (e.g. voltage vs energy) of which there is no current standard but would be good to work towards. I might have time next week to review/test the code more thoroughly, but otherwise looks good.

Tagging @jdickerson95 @paloha @volodymyr-masalitin for visibility

McHaillet:
Really good stuff! Thanks for making the PR @joelyeois

joelyeois:
Thanks for the feedback, everyone! Let me know if the energy/voltage needs adjusting.

Also, the current implementation assumes the entire 3D potential volume fits on the GPU. Future work would need to extend this to handle much larger 3D/4D tensors (e.g., full-resolution tomograms) that exceed GPU memory. Probably some form of CPU <-> GPU transfer but I'm not sure about the best way to implement this (if it's even needed).

McHaillet:
> Also, the current implementation assumes the entire 3D potential volume fits on the GPU. Future work would need to extend this to handle much larger 3D/4D tensors (e.g., full-resolution tomograms) that exceed GPU memory. Probably some form of CPU <-> GPU transfer but I'm not sure about the best way to implement this (if it's even needed).

Would it be possible to give atoms as input and sample them on the fly to a projection? I am thinking about multislice specifically. You could sample the atoms per slice. It would require some indexing into large sets of atoms of course

joelyeois:
You would also need to convolve each atom by its respective scattering factors on the fly I guess? Not sure how efficient this would be.

My current approach in my own simulator for cryoET data is to compute the entire scattering potential for 3D first, keep the volume on the CPU, and then iterate over batches of slices on the GPU so that the propagating 2D wavefunction stays on the GPU.

McHaillet:
> You would also need to convolve each atom by its respective scattering factors on the fly I guess? Not sure how efficient this would be.

You would, although you could do it in 2D in that case. But not sure if it would be more efficient.

McHaillet (approved):
PR looks really solid from my part, so let's get it in. We can always iterate later on if @paloha and @volodymyr-masalitin have suggestions. 

Concerning energy vs. voltage, my preference goes to the latter, since we also call it voltage in torch-ctf.

mgiammar:
+1 on voltage over energy

joelyeois:
Renamed all instances of `energy`/`energy_kev` to `voltage` to match torch-ctf's convention. Thanks for the feedback!

jdickerson95:
I'm happy to merge this once the tests are passing.
@joelyeois It looks like we have test_projection.py in torch-fit-in-map as well. Maybe rename it here to test_projection_scattering.py

joelyeois:
Renamed to `test_projection_scattering.py` in c0fd2f1 — thanks!

PR #98, opened 2026-07-16
Migrate torch-tilt-series into the monorepo (packages/primitives)

davidetorre99:
Replaces the outdated packages/wip/torch-tilt-series placeholder with the current package: tilt series geometry, named coordinate-space transforms (sample/levelled/tomogram/microscope/detector), and point projection.

PR #99, opened 2026-07-16
Migrate torch reconstruct tomogram

davidetorre99:
Depends on #98: this branch references torch-tilt-series as a workspace member. Diff includes that PR's commits until then, will shrink to just this package once it's merged and this branch is rebased.


packages/algorithms/torch-reconstruct-tomogram/src/torch_reconstruct_tomogram/projection.py
```diff
@@ -0,0 +1,85 @@
...
+    normalize_on_central_crop,
+)
+
+LocalShiftFn = Callable[[torch.Tensor], torch.Tensor]
+
+
+def project_points(
+    tilt_series: TiltSeries,
+    points_zyx: torch.Tensor,
+    local_shifts: LocalShiftFn | None = None,
```
McHaillet: Local shifts should be attached to the TiltSeries, right?

McHaillet (commented):
I had one comment when reading through it. Besides that it looks really nice.

packages/algorithms/torch-reconstruct-tomogram/src/torch_reconstruct_tomogram/projection.py
davidetorre99: That's actually an interesting point worth discussing. Right now, localshift is called during reconstruction so that you can change local shifts at every reconstruction without changing the TiltSeries object itself, and it's essentially used only in reconstruction and project_points.  Would you rather have it as a sample attribute (within the canonical sample space) in TiltSeries? I actually agree with you that it would probably fit better within the TiltSeries object, but I was originally thinking it would mainly be used for reconstruction, so it could just be a function to call and add to the original TiltSeries model.

alisterburt: I think the idea of being generic over local shift functions is in theory nice but in practice we need a single model on the object itself (which can then be serialized)

ideally this will be
- input/output in angstroms relative to image/volume center
- based on torch-cubic-spline-grids 
- provide methods for fitting the model based on observations from any other model (enabling import of local alignments from other software)

none of this functionality needs to  block getting what we've got in, can come with iteration :-)

davidetorre99: sounds great! Then for this PR I would just put local shift as part of the canonical sample in the TiltSeries object and then we'll build on top of that

McHaillet: Sounds great!

davidetorre99: Sorry if I keep iterating on this, but I think we can already consider two kinds of local shifts: a 3D (sample) shift and a 2D (per-tilt) shift. If we have everything in torch-tilt-series then TiltSeries.project_points does this sequence of steps:

3D point (tomogram space)
   ↓ tomo2sample
3D point (sample space)
   ↓ [local_shifts # 1 corrects the point here, still in 3D]
   ↓ project through geometry (rotate/tilt/etc.)
2D point (one per tilt image)
   ↓ [local_shifts # 2 corrects the point here, now in 2D]
final 2D position

So within TiltSeries we can have 
- `local_shifts` (3D, sample-space) — corrects a point's true 3D position before per-tilt projection. Models specimen structure (e.g. local warping).
- `local_shifts_2d` (2D, per-tilt) — corrects the projected 2D position, independently per tilt. 

How does this sound? Does it make sense? Right now they are both implemented but essentially local_shift_2D is what we have in torch-reconstruct-tomogram (and will now move to torch-tilt-series).
Another option is to already rename `local_shifts` (3D) something like `local_deformation`. I guess it's more a question about naming..

McHaillet: Sorry for the late reply. Yep, I agree, this sounds very good!

McHaillet: Ill wait for you to update that and then I will merge it in

McHaillet:
Thanks for updating @davidetorre99, looks good to me!

PR #102, opened 2026-07-27
(torch-so3) Implement multi-resolution S2 grid based on HEALPix

mgiammar:
Adds a new class to package `torch-so3` for defining a multi-resolution s2 grid using HEALPix/healpy functions. Also includes a demonstration notebook for the added API and functionality.


jdickerson95:
There is some redundancy with the healpix_sectored_base_grid() in the base_s2_grid.
I would suggest rewriting that function as a wrapper around the new class.

mgiammar:
@jdickerson95 good catch updated the base s2 grid to leverage the new code. Passes unit tests, but lmk if it breaks any of your analysis scripts.

jdickerson95:
Would it be cleaner to have some extra convenience functions in the class for constructing the grid from n_coarse + n_fine or theta_step, and then passing the grid into the healpix_sectored_base_grid?
We could also probably pass this grid into healpix_base_grid as well.

mgiammar:
Yes, the construction and bounds checking can be done by the multi-resolution grid class. Would make the sectored_base_grid method simpler.

I do like having the standalone healpix_base_grid_method since it is relatively simple, self-contained, and easily copyable.

packages/primitives/torch-so3/src/torch_so3/hierarchical_s2_grid.py
```diff
@@ -0,0 +1,386 @@
+"""Hierarchical S2 grid implementation based on healpy/healpix."""
+
+from dataclasses import dataclass
+
+import healpy as hp
```
jdickerson95: I think windows is now broken.

hierarchical_s2_grid.py does unconditional import healpy, and base_s2_grid always imports HierarchicalS2Grid. That pulls healpy into every torch_so3 import (uniform_base_grid, package __init__, etc.). Previously healpy was gated so non-HEALPix paths still worked on Windows.

jdickerson95 (commented):
Other comments:

- HierarchicalS2Grid not exported from torch_so3.__init__ but it is impiorted in nb.
- Tests have been added for sectored grid but no for the core HierarchicalS2Grid behaviour.

PR #105, opened 2026-08-05
feat(torch-refine-tilt-axis-angle): rewrite as a grid search on the Fourier sum of the stack

### LINKED ISSUE #94: tilt-axis angle determination using EMAN2 conventions
@alisterburt I was trying out some tilt-axis angle determination methods, and I found out that this one from EMAN2 is actually really easy and fast. I will make a PR to replace our current tilt axis angle calculation. (implemented here by Claude)

```python
import numpy as np
from scipy.fft import rfft2

def find_tilt_axis(stack: np.ndarray,
                    angle_step: float = 0.5,
                    min_radius_frac: float = 0.08,
                    max_radius_frac: float = 0.95,
                    refine: bool = True,
                    refine_range: float = 3.0,
                    refine_step: float = 0.1) -> float:
    """
    Determine the tilt-axis angle (0-180 deg) of an already ramp- and
    bandpass-filtered, translationally aligned tilt series.

    Handles rectangular images natively -- no cropping to square, so no
    image data is discarded. This matters because when ny != nx, the two
    Fourier axes have different bin spacings (1/ny cycles/pixel along rows,
    1/nx along columns), so a straight line at a given physical angle does
    NOT correspond to equal integer steps in (row, col) index space. Instead
    of taking radius steps in raw pixel index, this uses a normalized
    frequency rho in [0, 0.5) shared by both axes (same physical pixel size
    assumed in both directions) and converts to pixel-index coordinates
    separately for each axis (row index scaled by ny, column index scaled
    by nx). For a square image this reduces exactly to the earlier approach.

    Two-stage search:
      1. A coarse grid search over angle_step-sized steps across the full
         0-180 deg range (fast, robust, can't get stuck around the wrong
         peak).
      2. A finer grid search over refine_step-sized steps, restricted to
         +/- refine_range deg around the coarse optimum, to recover
         sub-angle_step precision without the cost of a fine search over
         the full range.

    Uses rfft2 rather than fft2: real-valued images guarantee
    F(-k) = conj(F(k)), so the full complex spectrum carries no information
    beyond its non-negative-frequency half, and the 0-180 deg search only
    ever needs one point per (angle, rho) rather than a +rho/-rho pair
    (which would just be a constant, angle-independent rescaling).
    """
    stack = np.asarray(stack)
    n, ny, nx = stack.shape

    # --- 1. Build a Hann taper (per-axis, handles ny != nx naturally) ---
    # This is what suppresses the FFT edge-artifact cross that comes from
    # the image not being periodic at its boundary.
    mask = np.outer(np.hanning(ny), np.hanning(nx))

    # --- 2. Coherent complex rfft sum across the stack ---
    F_sum = np.zeros((ny, nx // 2 + 1), dtype=complex)
    for img in stack:
        windowed = (img - img.mean()) * mask
        F_sum += rfft2(windowed)
    power_sum = np.abs(F_sum) ** 2

    # --- 3. Set up the shared normalized-frequency sample points ---
    # rho is a dimensionless fraction of Nyquist, shared by both axes; it is
    # converted to per-axis pixel-index coordinates (scaled by ny and nx
    # respectively) only at lookup time. The number of samples along the
    # line is set by the *longer* image dimension, since that axis offers
    # finer frequency-bin spacing (more distinguishable steps up to the same
    # physical Nyquist limit) -- sampling at only the shorter dimension's
    # density would waste the extra resolution the long axis provides.
    n_samples = max(ny, nx) // 2
    rhos = np.linspace(min_radius_frac * 0.5, max_radius_frac * 0.5, n_samples)

    def score(theta_deg: float) -> float:
        """
        Total power along the line through the origin at angle theta_deg,
        using only the stored (non-negative-frequency) half of the rfft
        spectrum: for directions with cos(theta) < 0, look up the
        conjugate-symmetric point F(-fy,-fx) = conj(F(fy,fx)) instead,
        which has the same magnitude.
        """
        rad = np.deg2rad(theta_deg)
        fy = rhos * np.sin(rad)
        fx = rhos * np.cos(rad)
        if np.cos(rad) < 0:
            fy, fx = -fy, -fx

        rows = np.round(fy * ny).astype(int) % ny
        cols = np.round(fx * nx).astype(int)
        valid = (cols >= 0) & (cols < power_sum.shape[1])
        return float(power_sum[rows[valid], cols[valid]].sum())

    # --- 4. Coarse grid search over 0-180 deg ---
    coarse_angles = np.arange(0, 180, angle_step)
    coarse_scores = np.array([score(a) for a in coarse_angles])
    best_angle = float(coarse_angles[np.argmax(coarse_scores)])

    if not refine:
        return best_angle

    # --- 5. Fine grid search around the coarse optimum ---
    fine_angles = np.arange(best_angle - refine_range,
                             best_angle + refine_range + refine_step,
                             refine_step)
    fine_scores = np.array([score(a) for a in fine_angles])
    best_angle = float(fine_angles[np.argmax(fine_scores)])

    return best_angle % 180.0
```

The relevant EMAN2 methods section for reference + their figure that shows it https://www.nature.com/articles/s41592-019-0591-8/figures/7:

To seed the iterative tilt series alignment, a coarse alignment is first performed. First, the unaligned tilt series is downsampled to 512 × 512 pixels, subjected to a real-space ramp filter and Fourier bandpass filter, and normalized to mean value of zero and standard deviation of one. A coarse alignment is then performed under a circular mask with soft Gaussian falloff. The alignment begins with the center tilt image (typically near untitled) and propagates sequentially in both directions. After the coarse translational alignment, the approximate tilt axis is identified by computing a coherent sum of the tilt series in Fourier space. The tilts will approximately share one Fourier common-line without any rotational alignment. This can be readily identified by simply performing a coherent sum after translational alignment and looking for the axis with maximum intensity (Supplementary Fig. [3](https://www.nature.com/articles/s41592-019-0591-8#Fig7)). Only angles 0–180 degrees are permitted in this process to ensure that no handedness flips occur between different tomograms within the same set. Although the handedness is thus consistent throughout the dataset, it is not necessarily correct, due to the 180-degree ambiguity in the tilt axis direction. If the correct orientation of the tilt axis in the images has already been determined for the microscope, it can be specified as a fixed parameter instead of performing a search.

McHaillet:
Closes #94

Before this ran a gradient descent optimization algorithm that would for each angle extract common lines from the 2D images. This was prone to end up in a local minimum, because its not a smooth landscape.

With this update: it sums the power spectra of the aligned tilt stack. It then search magnitudes around possible angles of the common line. Its all vectorized so its much faster. By default, it does a coarse grid search first in the range of [initial_axis - 90, initial_axis + 90] with a .5 degrees increment. It then refines the detected in the angle in the range of [current_best - 3, current_best + 3] with a 0.1 degrees increment.

I am also thinking if it would not be easier if we would just have one teamtomo.tilt_series_alignment module from which find_tilt_axis_angle and tiltxcorr could be imported. As well as some bandpass + edge taper + ramp subtraction, instead of having to do that both in tiltxcorr and in find_tilt_axis_angle. But I will make an issue about that after this to discuss further.

### DIFF
--- a/packages/algorithms/torch-refine-tilt-axis-angle/pyproject.toml
+++ b/packages/algorithms/torch-refine-tilt-axis-angle/pyproject.toml
@@ -45,11 +45,6 @@ classifiers = [
 # add your package dependencies here
 dependencies = [
     "torch",
-    "einops",
-    "torch-fourier-slice",
-    "torch-cubic-spline-grids",
-    "torch-affine-utils",
-    "torch-grid-utils",
 ]
 
 # https://peps.python.org/pep-0621/#dependencies-optional-dependencies
--- a/packages/algorithms/torch-refine-tilt-axis-angle/src/torch_refine_tilt_axis_angle/refine_tilt_axis_angle.py
+++ b/packages/algorithms/torch-refine-tilt-axis-angle/src/torch_refine_tilt_axis_angle/refine_tilt_axis_angle.py
@@ -1,112 +1,153 @@
 """Refine an initial tilt axis angle."""
 
-import einops
 import torch
-from scipy.optimize import minimize_scalar  # type: ignore[import-untyped]
-from torch_affine_utils.transforms_2d import R
-
-# teamtomo torch functionality
-from torch_fourier_slice import project_2d_to_1d  # type: ignore[import-untyped]
-from torch_grid_utils import circle  # type: ignore[import-untyped]
 
 
 def refine_tilt_axis_angle(
     tilt_series: torch.Tensor,
-    tilt_axis_angle: float = 0.0,
-    search_range: float = 10.0,
-    max_iter: int = 20,
+    tilt_axis_angle: float = 90.0,
+    coarse_angle_step: float = 0.5,
+    min_fraction_of_nyquist: float = 0.08,
+    max_fraction_of_nyquist: float = 0.95,
+    refine: bool = True,
+    refine_range: float = 3.0,
+    refine_angle_step: float = 0.1,
 ) -> float:
-    """Refine the tilt axis angle for electron tomography data.
+    """Find the tilt-axis angle of a tilt series from its common line.
 
-    Uses common line projections and scipy's Brent's method (bounded) to find
-    the optimal tilt axis angle that minimizes differences between projections
-    across the tilt series.
+    All images of a translationally-aligned tilt series share Fourier
+    information along a common line through the origin, perpendicular to
+    the tilt axis. Coherently summing the images' Fourier transforms makes
+    this line stand out as a ridge of high power, so the tilt-axis angle
+    can be recovered by finding the ridge's orientation via a two-stage
+    (coarse, then fine) angular grid search, evaluated as batched tensor
+    operations across all candidate angles at once.
+
+    Rectangular images are handled natively, via a normalized frequency
+    shared by both axes and converted to per-axis pixel indices only at
+    lookup time.
 
     Parameters
     ----------
     tilt_series : torch.Tensor
-        Tensor containing the tilt series images with shape [n_tilts, height, width].
-    tilt_axis_angle : float, default=0.0
-        Initial guess for the tilt axis angle in degrees.
-    search_range : float, default=10.0
-        Search range around the initial angle in degrees (±search_range).
-    max_iter : int, default=10
-        Maximum iterations for Brent's method optimizer.
+        Tensor containing the tilt series images with shape
+        `(n_tilts, h, w)`. Should already be translationally aligned and,
+        ideally, ramp- and bandpass-filtered. A pixel size of 10A is
+        recommended.
+    tilt_axis_angle : float, default=90.0
+        Initial guess for the tilt axis angle in degrees. The search range
+        is `[tilt_axis_angle - 90, tilt_axis_angle + 90]`. The default of
+        90.0 searches the full `[0, 180]` range.
+    coarse_angle_step : float, default=0.5
+        Step size for the coarse grid search (degrees).
+    min_fraction_of_nyquist : float, default=0.08
+        Minimum frequency radius searched, as a fraction of Nyquist.
+    max_fraction_of_nyquist : float, default=0.95
+        Maximum frequency radius searched, as a fraction of Nyquist.
+    refine : bool, default=True
+        Whether to run a finer search around the coarse optimum.
+    refine_range : float, default=3.0
+        Range around the coarse optimum searched during refinement
+        (degrees).
+    refine_angle_step : float, default=0.1
+        Step size for the fine grid search (degrees).
 
     Returns
     -------
     float
-        The optimized tilt axis angle in degrees.
-
-    Notes
-    -----
-    The function works by:
-    1. Projecting images perpendicular to the tilt axis to extract common lines
-    2. Comparing these projections across different tilts
-    3. Minimizing differences between projections using Brent's method
-
-    Common line projections are normalized and weighted according to a
-    projected spherical mask to emphasize regions of interest.
+        The tilt axis angle in degrees, in
+        `[tilt_axis_angle - 90, tilt_axis_angle + 90]`.
     """
-    n_tilts, h, w = tilt_series.shape
+    _, h, w = tilt_series.shape
     device = tilt_series.device
-    size = min(h, w)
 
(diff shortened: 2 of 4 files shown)


mgiammar:
I'll look over this today (if GitHub decides to become more stable), but generally think having a shared `teamtomo.tilt_series_alignment` would be useful if there's multiple functions depending on the same logic.

packages/algorithms/torch-refine-tilt-axis-angle/src/torch_refine_tilt_axis_angle/refine_tilt_axis_angle.py
```diff
@@ -1,112 +1,153 @@
...
+    cos, sin = torch.cos(rad), torch.sin(rad)
+    fy = sin[:, None] * rhos[None, :]  # (angle, rho)
+    fx = cos[:, None] * rhos[None, :]  # (angle, rho)
+    flip = (cos < 0)[:, None]
+    fy, fx = torch.where(flip, -fy, fy), torch.where(flip, -fx, fx)
+
+    rows = torch.round(fy * h).long() % h
+    cols = torch.round(fx * w).long()
+    valid = (cols >= 0) & (cols < power_sum.shape[-1])
+    values = power_sum[rows, cols.clamp(0, power_sum.shape[-1] - 1)]
```
mgiammar, line 152: As I understand, this doesn't do any interpolation along the line(s) defined by `angles_deg`. Would there be value in adding an interpolation stage across the 4 neighboring pixels for a point along the line?

May not be useful or too much effort, though. Mostly just curious.

packages/algorithms/torch-refine-tilt-axis-angle/src/torch_refine_tilt_axis_angle/refine_tilt_axis_angle.py
```diff
@@ -1,112 +1,153 @@
...
-    # use a spherical real-space alignment mask, quick way to get around #4
-    alignment_mask = circle(
-        radius=size // 3,
-        smoothing_radius=size // 6,
-        image_shape=(h, w),
+    # Hann taper: suppresses the FFT edge artifact from non-periodic boundaries.
+    mask = torch.outer(
+        torch.hann_window(h, periodic=False, device=device),
+        torch.hann_window(w, periodic=False, device=device),
+    )
```
mgiammar, line 68: Logic behind the hann window (as opposed to other windowing function)? Would you expect any result dependence upon changing the windowing function?

packages/algorithms/torch-refine-tilt-axis-angle/pyproject.toml
```diff
@@ -45,11 +45,6 @@ classifiers = [
 # add your package dependencies here
 dependencies = [
     "torch",
-    "einops",
-    "torch-fourier-slice",
-    "torch-cubic-spline-grids",
-    "torch-affine-utils",
-    "torch-grid-utils",
```
mgiammar, line 52: Love to see a reduction in cross-dependencies

packages/algorithms/torch-refine-tilt-axis-angle/src/torch_refine_tilt_axis_angle/refine_tilt_axis_angle.py
McHaillet: Hann windows are a common choice to prevent FFTing the discontinuities at the image border, but it might be a bit too aggressive. In torch-tiltxcorr I used a tukey window, which is an edge taper function that is rectangular with a cosine fall-off towards the edges. I think I will choose something consistent between the two, when I merge it into the teamtomo.tilt_series_alignment

packages/algorithms/torch-refine-tilt-axis-angle/src/torch_refine_tilt_axis_angle/refine_tilt_axis_angle.py
McHaillet: You're right it's just nearest neighbour. It might be better to do linear interpolation. I'd need to test.

McHaillet: I will leave it as is for now, I assume that for the line with most power nearest neighbour is sufficient.

PR #106, opened 2026-08-06
Experimental Mojo kernel backend for torch-fourier-slice

alisterburt:
Hi @mchaillet — would value your eyes on this one, especially the gridding correction section below, which turned up a bug that also affects the non-experimental code.

## What this is

A `torch_fourier_slice.experimental` namespace holding the same Fourier-slice operators as the rest of the package, with their compute kernels written in [Mojo](https://www.modular.com/mojo) and called from PyTorch. Everything is additive and opt-in: it lives under `experimental/`, needs the optional `mojo` extra, and `mojo_kernels_available()` reports whether the kernels built so callers can degrade gracefully. Nothing outside `packages/primitives/torch-fourier-slice/` changes except one `.gitignore` entry for the compile cache.

The pitch is that the numeric core is written **once** and lowered to CPU SIMD, NVIDIA PTX and Apple Metal from that single source — no `.cu` plus `.metal` plus a CPU fallback, and no wheel/CUDA-version matrix to publish, because the `.mojo` source compiles on first import for the machine it lands on. `experimental/README.md` is written for a torch user who hasn't written GPU kernels and explains the whole design.

## API

Two layers, both following the input tensor's device (CPU tensor → threaded CPU kernel, `mps`/`cuda` tensor → GPU kernel, output back on that device):

- **Real-space** — `project_3d_to_2d` / `backproject_2d_to_3d` (+ `_multivolume`), mirroring the package's own real-space functions. Real in, real out.
- **Fourier** — `extract_/insert_central_slices_rfft_3d` and `extract_/insert_central_line_rfft_{2d,3d}`, rfft in/out with DC at the origin.

Central **lines** are new: a line through the origin of a central slice is a line through the origin of the 3D transform, so a 1D central line is the degenerate central slice whose in-plane axis has collapsed to its DC row. It's posed by a bare direction vector rather than a rotation matrix, since rotating about the line's own axis is a gauge its values are blind to.

Everything is differentiable w.r.t. its data, its pose (rotation matrices or direction vectors), 2D/3D shifts, and insertion weights — the pose and shift gradients are dedicated backward kernels, all validated by finite differences.

The GPU path reads and writes the memory behind torch tensors in place (no host round-trip), and on CUDA enqueues on torch's current stream rather than forcing a device sync.

## ⚠️ Gridding correction — this one affects the main package too

Interpolating in Fourier space is a convolution with the interpolation kernel, which multiplies the **real-space** result by `K`, that kernel's continuous Fourier transform. So the correction is a **division** by `K`. Two things were wrong:

1. **Direction.** The forward path *multiplied* by `sinc²`, which applies the apodization a second time instead of removing it.
2. **`sinc²` is the linear kernel only.** For tricubic Catmull-Rom (Keys `a = -1/2`) the transform is `K(ν) = sinc³(ν)·(3·sinc(ν) − 2cos(πν))` — derived by integrating the kernel, and re-derived against numerical quadrature in the tests so it isn't a magic constant. Both kernels are separable, so the 3D correction is the outer product of the per-axis 1D transforms rather than a function of `|ν|`.

Measured against the analytic projection of an isotropic blob, whose projection is identical at every orientation and so is an exact ground truth (relative RMS, 64³ box):

| | no correction | `sinc²` | matched `K` |
|---|---|---|---|
| linear | 7.8e-4 | — | **1.6e-4** |
| cubic | 1.6e-5 | 7.9e-4 | **1.4e-5** |

**The same direction bug is in the non-experimental code**: `project_3d_to_2d_multichannel` and `project_2d_to_1d` pre-multiply by `sinc²`, and single-volume `project_3d_to_2d` applies no correction at all (it zeroes and restores the mean instead). `backproject_2d_to_3d` divides, which is correct. I deliberately did **not** touch any of it here — that's a numerical behaviour change to a released API and deserves its own PR/discussion. For scale, on that same blob the canonical projector measures 4.2e-3 where the experimental one gets 1.7e-4.

## Example

`examples/ribosome_reconstruction_demo.py` simulates a density map for the 80S ribosome (`4v6x`) from its RCSB entry with `ttsim3d`, projects it at random orientations, reconstructs from those projections, and scores by FSC against the input volume.

- Noise-free, 400 projections: FSC ≥ 0.5 out to Nyquist (6.0 Å), mean FSC 0.964 — really a check that the projector and backprojector are exact adjoints.
- At SNR 0.1, resolution tracks particle count: 100 → 34.9 Å, 400 → 25.6 Å, 1600 → 17.5 Å.

Uses only deps already in the `test` group. First run downloads and simulates (~30 s), then caches; after that it's a couple of seconds.

## Testing

127 passed, 0 skipped, on CPU and Apple/Metal; the CUDA path was validated separately on an A100. Two tests that needed an external comparison package (and so always skipped) were replaced with self-contained equivalents — insertion against the canonical teamtomo kernel, and cubic-vs-linear accuracy against an exact ground truth.

## Reviewing

The diff is large but most of it is Mojo kernels and tests. Suggested order: `experimental/README.md` for the design, then `_gridding.py`, `project.py` and `backproject.py` for the part with real numerical consequences.

https://claude.ai/code/session_012hUv8YSdo8Dr4XiQMJCr12


packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/README.md
```diff
@@ -0,0 +1,415 @@
...
+```
+
+The **Fourier** layer is the same operators with the transforms left to you —
+rfft in, rfft out, DC at the origin. Use it when you are already working in
+Fourier space and don't want a round trip per call:
+
+```python
+from torch_fourier_slice.experimental import extract_central_slices_rfft_3d
+
+volume_rfft = torch.fft.rfftn(torch.fft.fftshift(volume, dim=(-3, -2, -1)),
```
McHaillet, line 80: Should this actually not be an ifftshift? We move the center of the real space volume to the Fourier transform origin, not the other way around. 

Got sidetracked with this haha, awesome PR!

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/README.md
```diff
@@ -0,0 +1,415 @@
...
+                              dim=(-3, -2, -1))
+slices = extract_central_slices_rfft_3d(volume_rfft.contiguous(), rotations)
```
McHaillet, line 82: And the 'old' torch fourier slice had to also run an fftshift on the volume_rfft. I guess you now wrote it so that the interpolation kernels work on rffts with the DC at the origin, correct?

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/README.md
```diff
@@ -0,0 +1,415 @@
...
+projection direction. So:
+
+- **Forward (project):** for each 2D output pixel, rotate its frequency
+  coordinate into the volume, **sample+interpolate** the 3D rfft volume there,
+  apply any shift phase. This is a **gather** (each output reads a few inputs).
+- **Backward (backproject):** for each 2D input pixel, rotate into the volume and
+  **splat+accumulate** its value into the nearby voxels. This is a **scatter**
+  (each input writes a few outputs, hence atomics).
+
+Everything is done in **rfft layout with DC at the origin** (real volume →
```
McHaillet, line 104: okay nvm my previous comment

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/README.md
```diff
@@ -0,0 +1,415 @@
...
+`TileTensor` view `[d, h, w, 2]` over the zero-copy pointer (`_load_c2` /
+`_store_c2`); atomic scatter accumulation stays on raw pointers.
+
+---
+
+## 11. Differentiability
+
+Every op is **fully differentiable**: the extraction w.r.t. `reconstruction`,
+`rotations` (or `directions`), `shifts`; the insertion w.r.t. `projections`,
+`weights`, `rotations` (or `directions`), `shifts`.
```
McHaillet, line 380: More of a meta question: should projection (aka extraction) not also have an option for weights?

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/_mojo/fourier_slice_kernels.mojo
```diff
@@ -0,0 +1,1837 @@
...
+    )
+    var bv = Int(py=rec_obj.shape[0])
+
+    @parameter
+    def worker[interp: Int](vp: Int):
+        var i_bv = vp // p.bp
+        var i_bp = vp % p.bp
+        for y in range(p.proj_sidelength):
+            for x in range(p.proj_sidelength_half()):
+                _project_pixel[interp](
```
McHaillet, line 277: So _project_pixel is the one for all function that determines the interpolation and is the same for cpu/gpu, correct? And in the CPU each thread just loops over all the pixels in the image.

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/_mojo/fourier_slice_kernels.mojo
```diff
@@ -0,0 +1,1837 @@
...
+                    rec, rot, shifts_2d, shifts_3d, proj, i_bv, i_bp, y, x, p
+                )
+
+    if (
+        p.interp == CUBIC
+    ):  # runtime interp code -> pick the comptime specialisation once
+        parallelize[worker[CUBIC]](bv * p.bp, num_physical_cores())
```
McHaillet, line 284: I am just trying to understand how parallelization works here
* parallelize seems to be a mojo standard library function
* worker is the function that we parallelize
* bv * b.pb = number of volume batches * number of rotation batches (i.e. total number of tasks)
* in worker i_bv and i_bp index the volume batch and rotation batch

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/_mojo/fourier_slice_kernels.mojo
```diff
@@ -0,0 +1,1837 @@
...
+    address as the trailing element (0 to enqueue on the context's own stream;
+    see `_launch_project`). The tensor objects are used only for their shapes.
+    `proj` is pre-zeroed on the device by the caller (radius-cut pixels stay 0,
+    matching the CPU path). We sync the context only on the own-stream path.
+    """
+    var p = _forward_params(
+        rec_obj, rot_obj, shifts_2d_obj, shifts_3d_obj, proj_obj, params_obj
+    )
+    var bv = Int(py=rec_obj.shape[0])
+    var total = bv * p.bp * p.proj_sidelength * p.proj_sidelength_half()
```
McHaillet, line 927: and on the gpu each thread just handles a single pixel, so this is (sort of) the total number of tasks

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/_mojo/_pixel.mojo
```diff
@@ -0,0 +1,153 @@
...
+    p: FourierSliceParams,
+):
+    """Gather one output rfft pixel (i_bv, i_bp, y, x) of the central slice."""
+    var coord_y = _fourier_coord(y, p.proj_sidelength)
+    var coord_x = Float32(x)
+    if coord_y * coord_y + coord_x * coord_x > p.radius_cutoff_sq:
+        return  # outside the cutoff: leave the (pre-zeroed) output at 0
+
+    var rb = 0 if p.bv_rot == 1 else i_bv
+    var sx = coord_x * p.oversampling
```
McHaillet, line 51: here there is an oversampling parameter, but in the actual function signatures (i.e. backproject / project) I don't think I saw this parameter.

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/_mojo/_pose_grad.mojo
```diff
@@ -0,0 +1,452 @@
+"""Per-pixel backward ops for gradients w.r.t. rotations, shifts_2d and weights.
```
McHaillet, line 1: Why is this file called _pose_grad.mojo? Shouldn't it be _pixel_grad.mojo like the other files?

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/_autograd.py
```diff
@@ -0,0 +1,687 @@
...
+    """
+    while grad.dim() > target.dim():
+        grad = grad.sum(0)
+    for dim in range(grad.dim()):
+        if target.shape[dim] == 1 and grad.shape[dim] != 1:
+            grad = grad.sum(dim, keepdim=True)
+    return grad.reshape(target.shape).to(device=target.device, dtype=target.dtype)
+
+
+def _symmetrise_kx0_plane(grad_volume: torch.Tensor) -> torch.Tensor:
```
McHaillet, line 61: I get these are needed to ensure the symmetry on the plane (and the function below the column). But I thought this could be handled in the kernels.

McHaillet: Can you give a brief explanation on how the Friedel symmetry is now handled? Since we had quite some trouble with it before, I am just trying to understand how its dealt with in the kernels.

McHaillet:
@alisterburt thanks for taking the time to put it into a PR, really cool stuff!

I added some initial comments which are mainly me trying to understand how its built up ;) Overall it looks very good, but its just a lot of code to parse

alisterburt:
Great to have eyes on it! Thanks for following up

pixel grad is grad wrt pixel data
pose grad is grad wrt pose variables so you can optimize either

Yes you're right about parallelization and that core function being reused across CPU/GPU and that those calculations are defining number of tasks :-) 

Yes oversampling is not exposed at the higher level

You're likely right about the fftshift->ifftshift thing (I whipped that final project/backproject python interface together far too quickly late last night)

I tried to clean the code up a bunch but there is likely LLM slop in various corners

Can I hand this off to you to get over the line? It's been working well in my hands and would love to see it merged but only if the quality is up to your standards 🙂

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/README.md
alisterburt: Erm, don't think so, what's your thinking?

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/_mojo/fourier_slice_kernels.mojo
alisterburt: Yep! Mojos support for multi threading is limited right now any this is the only way to access it

alisterburt (commented):
That huge Python module should be split up and cleaned up for sure, what a monster

jdickerson95:
Love to see this! I can give it a try in a couple of days.
In terms of the gridding correction, I agree it should be a separate PR. We use the extraction and not the projection, so we would not be impacted by numerical changes.

alisterburt:
totally fine with this being split up/modified in whatever way - could I leave this mostly in your guys hands? Can't commit much time right now

mgiammar:
Very cool will definitely try this out (even just to see if there's a throughput boost on slice extraction / projection).

Agree with @jdickerson95 that the gridding correction is independent and should be separated.

mgiammar:
@alisterburt  Might find time in two weeks, still backed up trying to get a paper submitted. But happy to test since this _could_ accelerate portions of that pipeline.

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/README.md
McHaillet: I thought for symmetry, but I guess it's a moot point. Forward projection is well-defined so doesn't require weighting, while reconstruction is inherently ill-posed.

packages/primitives/torch-fourier-slice/src/torch_fourier_slice/experimental/_mojo/fourier_slice_kernels.mojo
McHaillet: For our use case totally fine I think

McHaillet:
@mgiammar and @jdickerson95, this is all currently in an experimental folder of torch-fourier-slice, so your current calls anyway won't be impacted. From reading the gridding correction code for the mojo path, this does seem to be the proper way to do it. 

```
def _kernel_transform(frequency: torch.Tensor, interpolation: str) -> torch.Tensor:
    """Continuous Fourier transform of the interpolation kernel, ``K(v)``.

    ``frequency`` is a normalised real-space coordinate in ``[-0.5, 0.5)``. Both
    forms are written via :func:`torch.sinc` so they need no division and are
    exactly 1 at the origin.
    """
    sinc = torch.sinc(frequency)
    if interpolation == "linear":
        return sinc**2
    if interpolation == "cubic":
        return sinc**3 * (3 * sinc - 2 * torch.cos(torch.pi * frequency))
    raise ValueError(
        f"interpolation must be one of {sorted(_INTERPOLATIONS)}, got {interpolation!r}"
    )


def gridding_correction(
    sidelength: int,
    interpolation: str,
    device: torch.device | str | None = None,
) -> torch.Tensor:
    """Cubic ``(d, d, d)`` de-apodization volume for ``interpolation``.

    Divide a centered real-space volume by this to undo the apodization the
    interpolation kernel imposes. Built as an outer product of the per-axis 1D
    transform, so it never materialises an intermediate coordinate grid.
    """
    frequency = torch.fft.fftshift(torch.fft.fftfreq(sidelength, device=device))
    k = _kernel_transform(frequency, interpolation)
    return k[:, None, None] * k[None, :, None] * k[None, None, :]
```

McHaillet:
> You're likely right about the fftshift->ifftshift thing (I whipped that final project/backproject python interface together far too quickly late last night)

The rest of torch-fourier-slice also has fftshift first and ifftshift after so I'm guessing claude stuck to that. It doesn't even matter for even sized boxes of course so it's usually not a problem. But for correctsness sake it should be the other way around.

mgiammar:
> @mgiammar and @jdickerson95, this is all currently in an experimental folder of torch-fourier-slice, so your current calls anyway won't be impacted. From reading the gridding correction code for the mojo path, this does seem to be the proper way to do it.

I remember getting this same result a few months back when dealing with different slice extraction/insertion bugs. Grid correction should be applied in real-space and separable. Will take a look at #107

McHaillet:
Yes, I dug further into it and you're right: it should be separable. I made PR #107 to address it in the main torch-fourier-slice code, and after that we can also update it here.

McHaillet:
Hey @alisterburt, I ran the code and noticed on my laptop that cuda was slower than cpu. I pushed two commits that force atomic operations to be FLEXIBLE which greatly improves cuda performance on my laptop. Claude dug up in the mojo source that atomic by default happen sequentially on cuda (while they are flexible on mps), and that just forces all the GPU threads to wait for each other when they are inserting around the contended region close to the DC.

Benchmark against torch-projectors for `insert_central_slices_rfft_3d`:

| n particles | Mojo CUDA (before) | Mojo CUDA (after) | speedup | torch-projectors CUDA (reference) |
|---|---|---|---|---|
| 10 | 57.9 ms | 0.7 ms | 82x | 0.4 ms |
| 100 | 614.3 ms | 4.6 ms | 133x | 4.2 ms |
| 400 | 2610.8 ms | 20.3 ms | 128x | 17.6 ms |

This same fix was also applied for all the gradient code but it doesnt fully solve the speed:

| n particles | GPU backward (before) | GPU backward (after) | speedup | CPU backward (reference) |
|---|---|---|---|---|
| 10 | 297.9 ms | 62.4 ms | 4.8x | 38.1 ms |
| 100 | 3276.1 ms | 527.8 ms | 6.2x | 84.4 ms |
| 400 | 13286.3 ms | 2047.1 ms | 6.5x | 194.7 ms |

Real speedup (5-6.5x), but this does **not** close the gap to CPU: GPU is still ~10x slower at n=400. We might be able to solve it like this:

The gradient-accumulation kernels have a fundamentally worse contention pattern than the forward scatter, and relaxed ordering alone can't fix it. In the forward scatter, only pixels landing near the DC/low-frequency region collide across differently-oriented slices. In `_pose_grad.mojo` every pixel belonging to a given pose atomically adds into the same 9 (rotation) or 2-3 (shift) floats, unconditionally — for a 128x65-pixel slice, that's up to ~8,320 atomic operations into the same handful of addresses for one pose, with no local accumulation before hitting global memory.

Compare to `torch-projectors`' CUDA backward kernel (`backprojection_2d_to_3d_kernels.cu:963-997`), which does a proper block-level reduction: each of the 256 threads in a block (one block per pose) accumulates its own slice of that pose's pixels into a private shared-memory slot using plain non-atomic `+=` (zero contention during the entire per-pixel loop, since only that thread ever touches its own slot). Only after `__syncthreads()` does thread 0 sum the 256 partials and issue 9 atomics for rotation + 2 for shift — 11 total, once per pose — regardless of how many pixels that pose has:

```cpp
__syncthreads();
if (tid == 0) {
    float total_rot_grad[9] = {0.0f, ...};
    for (int t = 0; t < blockDim.x; ++t)
        for (int i = 0; i < 9; i++)
            total_rot_grad[i] += local_rot_grad[t][i];   // serial reduction, one thread
    for (int i = 0; i < 9; i++)
        atomic_add_real(&grad_rotations[global_rot_idx_base + i], total_rot_grad[i]);  // 9 atomics, once
}
```

alisterburt:
Ooh good digging, I vaguely checked perf for forward projection but never backprojection/grads. 

Any ideas where the remaining perf gap comes from? I've not done this myself but apparently this is a good intro to profiling torch GPU shenanigans https://youtu.be/LuhJEEJQgUM?is=ytyIXwl3HWb1FFjy

We should definitely add the block local reduction for the gradients

If we really care about backprojection perf we can imagine doing local accumulation of the low freq components too, should probably come in a follow up rather than this initial PR though

alisterburt:
Oh, and great job figuring out the FLEXIBLE thing, I haven't heard of this! Will do some reading :-)

alisterburt:
I see you also have a different solution from torch-projectors for removing write contention, that also looks great :-)

McHaillet:
> Any ideas where the remaining perf gap comes from? I've not done this myself but apparently this is a good intro to profiling torch GPU shenanigans https://youtu.be/LuhJEEJQgUM?is=ytyIXwl3HWb1FFjy

You mean compared to torch-projectors? It could be some block optimizations, but I am not entirely sure. With 20 ms vs 17 ms, it's not something major for sure.

McHaillet:
> We should definitely add the block local reduction for the gradients

I think that can already go in this PR!

jdickerson95:
@alisterburt @McHaillet @mgiammar any reason not to merge this?

mgiammar:
I think it should be fine to include in v0.6.0 release. Changes can be made from there.

alisterburt:
Nothing blocking

If someone could add an issue to track adding the warp(block?) local reductions to all of the insertion paths that would be great

alisterburt:
@mgiammar thanks for updating for 1.0 support

PR #107, opened 2026-08-10
fix(torch-fourier-slice): use separable sinc^2 for linear interpolation correction

### LINKED ISSUE #65: (torch-fourier-slice) Rationale behind multiplying by sinc2 grid in `project_3d_to_2d`
@alisterburt I remember you having some rationale (which I can't seem to remember) behind the sinc-squared grid multiplication during slice extraction:

https://github.com/teamtomo/torch-fourier-slice/blob/b53a76e0ae56e8086a36d929b6bd3ca4aeb149f1/src/torch_fourier_slice/project.py#L57-L65

I'm trying to re-update my mental model of how `torch-fourier-slice` is operating since there's some downstream computations whose numerical stability depends somewhat on having accurate projections. Would you mind re-explaining the rationale?

McHaillet:
@jdickerson95, @mgiammar, this should be the correct apodization correction from the discussion in #65

## Summary

Fixes the gridding correction applied around slice extraction/insertion in `torch-fourier-slice`. Closes #65.

## Why separable, not radial

Trilinear/bilinear interpolation is implemented as independent 1D linear interpolation along each axis in turn (see the corner-weighted lerp in `slice_extraction`/`slice_insertion`). Expanding the nested lerps shows every corner's weight factors as a product of per-axis terms:

```
w(x, y, z) = wx(fx) · wy(fy) · wz(fz),   w0(t) = 1-t,  w1(t) = t
```

i.e. the interpolation kernel in Fourier space is `tent(kx)·tent(ky)·tent(kz)` — a cube-shaped, axis-aligned kernel, not a radially symmetric cone. Its real-space correction is therefore also separable:

```
IFT[tent(k)] = sinc²(x)   (1D triangle/rect Fourier pair)
=> correction = sinc(x)² · sinc(y)² · sinc(z)²
```

The previous code instead computed `sinc(|r|)²` using the Euclidean norm of the grid — the correct correction only in 1D, or exactly on the three coordinate axes where the other components vanish. Off-axis it diverges increasingly with distance (worst along the cube diagonals), which shows up as the halo/ringing artifacts reported in #65.

### DIFF
--- a/packages/primitives/torch-fourier-slice/src/torch_fourier_slice/backproject.py
+++ b/packages/primitives/torch-fourier-slice/src/torch_fourier_slice/backproject.py
@@ -100,11 +100,14 @@ def backproject_2d_to_3d(
     dft = torch.fft.irfftn(dft, dim=(-3, -2, -1))
     dft = torch.fft.ifftshift(dft, dim=(-3, -2, -1))  # center in real space
 
-    # correct for convolution with linear interpolation kernel
+    # correct for convolution with linear interpolation kernel. The kernel is
+    # separable per-axis, so the correction is a product of 1D sincs, not
+    # the sinc of the radial frequency. See issue #65 for more info.
     grid = fftfreq_grid(
-        image_shape=dft.shape, rfft=False, fftshift=True, norm=True, device=dft.device
+        image_shape=dft.shape, rfft=False, fftshift=True, norm=False, device=dft.device
     )
-    dft = dft / torch.sinc(grid) ** 2
+    sinc = torch.sinc(grid[..., 0]) * torch.sinc(grid[..., 1]) * torch.sinc(grid[..., 2])
+    dft = dft / sinc**2
 
     # unpad
     if pad_factor > 1.0:
@@ -205,15 +208,18 @@ def backproject_2d_to_3d_multichannel(
     dft = torch.fft.irfftn(dft, dim=(-3, -2, -1))
     dft = torch.fft.ifftshift(dft, dim=(-3, -2, -1))  # center in real space
 
-    # correct for convolution with linear interpolation kernel
+    # correct for convolution with linear interpolation kernel. The kernel is
+    # separable per-axis, so the correction is a product of 1D sincs, not
+    # the sinc of the radial frequency. See issue #65 for more info.
     grid = fftfreq_grid(
         image_shape=dft.shape[-3:],
         rfft=False,
         fftshift=True,
-        norm=True,
+        norm=False,
         device=dft.device,
     )
-    dft = dft / torch.sinc(grid) ** 2
+    sinc = torch.sinc(grid[..., 0]) * torch.sinc(grid[..., 1]) * torch.sinc(grid[..., 2])
+    dft = dft / sinc**2
 
     # unpad
     if pad_factor > 1.0:
--- a/packages/primitives/torch-fourier-slice/src/torch_fourier_slice/project.py
+++ b/packages/primitives/torch-fourier-slice/src/torch_fourier_slice/project.py
@@ -9,7 +9,7 @@ from .slice_extraction import (
     transform_slice_2d,
     transform_slice_2d_multichannel,
 )
-from .volume_utils import compute_cube_face_averages, compute_square_edge_averages
+from .volume_utils import compute_cube_face_averages
 
 
 def project_3d_to_2d(
@@ -75,13 +75,11 @@ def project_3d_to_2d(
     if pad_factor < 1.0:
         raise ValueError("pad_factor must be >= 1.0")
 
-    edge_value = compute_cube_face_averages(volume, n=4)
-    volume = volume - edge_value
-
-    # Apply zero padding to the nearest integer matching the desired pad_factor
+    # Apply edge padding to the nearest integer matching the desired pad_factor
     if pad_factor > 1.0:
         p = int((w * (pad_factor - 1.0)) // 2)
-        volume = F.pad(volume, pad=[p] * 6, mode="constant", value=0.0)
+        edge_value = compute_cube_face_averages(volume, n=4)  # 4 is arbitrary
+        volume = F.pad(volume, pad=[p] * 6, mode="constant", value=edge_value)
 
     volume_shape = tuple(volume.shape[-3:])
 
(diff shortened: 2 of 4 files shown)


mgiammar:
Playing around with this in a demo notebook right now to verify correctness and there are some caveats and interactions with the projection portion of the code to be aware of:

- Volume density should exist within roughly central 35-40% of volume, if `pad_factor=1.0`, scaling with `pad_factor` (i.e. setting real-space zero-padding to 2x means can use 70-80% of volume. Derived from where 1/sinc^2(f)=1.1 which gives f=0.17 and based on the relative distance (d) from diameter of particle (D) is f=D/d so approx 17% on each side.
- Interacts with volume mean too since it gets scaled by the sinc^2 grid, need to adjust how the volume mean is being calculated and added back in for the `project_*` methods. Don't know if this affects backprojection.

Still though applying the correct gridding correction is giving about ~10% improvements in mae on some ground-truth Gaussian blobs (which are less sharp than typical for our volumes).

McHaillet:
Thanks for the quick review and fixing the mean! 

I'll have a think about the density occupation tomorrow

jdickerson95:
Should we not be restoring the mean of the volume before the division by sinc?
So do something like
calc mean of mean volume
subtract mean so division does not affect mean (stays zero)
divide by sinc
FFT and DC 0
Add back original mean.

I also agree with @mgiammar, the sinc will be overcorrecting at the boundaries. Should we document the difference padding makes somewhere in the package so users can see what difference the parameter makes?

McHaillet:
> Volume density should exist within roughly central 35-40% of volume, if pad_factor=1.0, scaling with pad_factor (i.e. setting real-space zero-padding to 2x means can use 70-80% of volume. Derived from where 1/sinc^2(f)=1.1 which gives f=0.17 and based on the relative distance (d) from diameter of particle (D) is f=D/d so approx 17% on each side.

As far as I know, yes, we always need to pad, but it's good to have it clear till what percentage the volume is useful. That should be added to the docs. It also matches my experience when I looked at the padding factor, 2x seems fine for reconstructing particles, for tomograms I need to set it to 3x for a reconstruction without boundary artifacts. I will add it to the PR. But, this is just inherent to the Fourier gridding algorithm and not so much to what the right real space correction is. 

> Still though applying the correct gridding correction is giving about ~10% improvements in mae on some ground-truth Gaussian blobs (which are less sharp than typical for our volumes).

Thanks for confirming, it should be the correct 'correction', because the interpolation is also separable per dimension.

McHaillet:
> calc mean of mean volume
> subtract mean so division does not affect mean (stays zero)
> divide by sinc
> FFT and DC 0
> Add back original mean.

If we DC 0 we are anyway removing the mean, right?

jdickerson95:
DC 0 should will set mean zero. I was just keeping it in case the sinc division changed the mean a little.
The main point is I think the mean should be restored from before sinc division

McHaillet:
Ah, I see what you mean. Also didn't fully read what the mean was used for, but now I get it.

I agree with you, but I think we should even calculate the mean before padding. Because right now we are storing `d,h,w` of the original volume, then padding, then calculating the mean. While at the end, we unpad, and only then add `d * mean`.

mgiammar:
> > calc mean of mean volume
> > subtract mean so division does not affect mean (stays zero)
> > divide by sinc
> > FFT and DC 0
> > Add back original mean.
> 
> If we DC 0 we are anyway removing the mean, right?

What's currently done (for slice extraction) is mean of original volume is computed, then after FFT, DC is set to zero. Upon transformation back into real-space, the appropriately scaled mean is added to the projections.

Circumvents the following issues:

- Strong low-res artifacts for bins around strong DC 0 pulling density during interpolation.
- Projecting uniform mean even when original cube is rotated off-axis (otherwise edges get clipped).
- Proper scaling of mean for a padded volume (and here sinc^2 scaled volume).

As far as I know, similar computations are not being done the slice insertion codepath @McHaillet

jdickerson95:
@mgiammar I don't think the mean of the original volume is computed:
```

    volume = volume / sinc**2

    # Compute the mean from the padded *and* corrected volume
    volume_mean = volume.mean()
```

This is the mean after padding and division by sinc. Agree with @McHaillet it should be the mean before padding as well

mgiammar:
Been thinking about how to properly account for mean and volume background (and testing implementations) in conjunction with the sinc^2 correction. The volume correction helps with the DC=0 bin interpolation artifacts, and the more impactful item now is the periodic boundary assumptions for the FFT and projecting the correct background regardless of upsampling factors.

Discussion of mean and background subtraction, and how to account for it:

> Assuming we have a cubic volume where structure density is only in some central spherical area, and assuming the structure is arbitrarily oriented, we want to model a constant projected background. If the near edge values of the volume model this background well and are constant, then the projected background should be `edge_values * d` where `d` is the depth of the volume.
>
> This background _should not_ be corrected for by some interpolation kernel since it is constant. Similarly, this should be modeled from the original volume, not the padded volume.
>
> Before the FFT and interpolation stage, the background should be zero (and padded with zeros) to obey periodic boundary conditions for the FFT and because otherwise a shadow of a cube would be projected. Only surrounding volume values of zero will project to zero in the image. Centering the volume by mean subtraction and/or setting DC to zero will violate this fact.

Fourier slice extraction should therefore be

1. Compute volume depth and average edge values, assuming a constant background near the edge.
2. Subtract off average edge value from volume
3. Zero pad, if requested
4. Do FFT and interpolation stages as normal
5. Add back average edge value to projections scaled by volume depth.

jdickerson95:
I like having the background edges at zero to prevent discontinuities with zero padding, but I have some questions.

1) How much does compute_cube_face_averages n parameter affect things / why is 4 default?
2) The math is much cleaner without the DC=0. I was assuming this was to prevent interpolation artefacts for cases where the DC is much higher than everything else.

McHaillet:
I feel like we're getting a bit side-tracked. Just to recap: I think all agree on the updated sinc2 correction.

Can you remind me why we need to change the mean calculation? My assumption is its because of zero padding causing discontinuities, or am I overlooking something? And why is the simple fix of moving the mean calculation to before the padding step not sufficient? 

I am also wondering if this is something that should be done inside this function or exposed to the user. Because we could also update the code to accept a value that is used for padding (which seems like a more general fix). That way code that requires it, can calculate the edge value in whatever way they like and supply it as the padding value.

> As far as I know, similar computations are not being done the slice insertion codepath @McHaillet

Definitely not, I also wouldn't know what sort of approach would mirror the inverse of this.

McHaillet:
My suggestion concerning mean/edges would be:

* in this PR only move the current mean calculation to before the padding step, because its just wrong otherwise
* leave the 'edge mean calculation' / 'constant padding value input' to another PR

mgiammar:
@McHaillet works for me. Will revert changes and open them in another PR

mgiammar:
Reverted changes, as requested. Note I had to lower the FSC cutoff from `0.99` to `0.98` on the projection-backprojection cycle unit test to get one of the volumes to pass.

Happy with the code.

mgiammar:
@McHaillet removed the mean computation (both setting DC to zero and tracking the mean) from the projection code so that unit tests pass. Discussed with Josh in the lab today and I don't think is a way to both account for the edge artifacts and the DC component.

Will follow up and correct in a separate PR.

McHaillet:
Merging in then!

PR #109, opened 2026-08-10
fix torch-fourier-slice for odd sized dimensions

McHaillet:
@mgiammar this is also a things I noticed when looking at PR #106, the fftshift and ifftshift should be swapped. It's only relevant for odd sized dimensions, which probably not a lot of people use, but it still seems important to get it correct.

Adds two fixes:
* supplies the s (shape) argument for irfftn which is needed to get the original shape back
* swaps the use off ifftshift/fftshift in real space. Shifting before taking the Fourier transform is needed to correctly place the box center at the origin. This is exactly what ifftshift (in contrast to fftshift) is intended for. So correct use is to ifftshift before and fftshift after. Note that this only matter for odd sized dimensions because the ifftshift/fftshift are identical for even sized dimensions.

I added tests that fail on main and succeed after the fixes in this branch.

### DIFF
--- a/packages/primitives/torch-fourier-slice/src/torch_fourier_slice/backproject.py
+++ b/packages/primitives/torch-fourier-slice/src/torch_fourier_slice/backproject.py
@@ -72,7 +72,7 @@ def backproject_2d_to_3d(
     volume_shape = (w, w, w)
 
     # calculate DFTs of images
-    images = torch.fft.fftshift(images, dim=(-2, -1))  # volume center to array origin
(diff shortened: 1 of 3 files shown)


mgiammar:
Makes sense to me since the volume/image center should be placed at (0, 0). Have not been working with odd-sized volumes/images so would never have encountered the bug. Code looks good to me!

mgiammar (approved):

McHaillet:
The other option is to error out on odd dimensions, but if it works

PR #111, opened 2026-08-12
fix: evaluate Zernike aberrations against physical frequency, not normalized rho

joelyeois:
I was trying to verify the different aberration terms between cryoSPARC and torch-ctf, and through Claude I came across this issue.

It looks like the Zernike radial coordinate gets normalized to `[0, 1]`, but the coefficients going into it are calibrated against physical frequency in cycles/Å. So beam-tilt coma comes out scaled by `1 / rho_max**3`, which varies with box size and pixel size. The line is in `_ctf_core.py`:

```python
rho, theta = fftfreq_grid_polar(fft_freq_grid)   # normalize_rho defaults to True
```

Is this accidental? Two things make me think it might be: `_build_freq_grid`'s docstring already says `rho : ... in cycles/Angstrom`, and `beam_tilt_to_zernike_coeffs` builds its prefactor as `2 * C.pi * Cs * lam**2` with both in Å — that's a rad·Å³ coefficient, so it wants a physical `k**3`. Those two can't both be right as they stand.

RELION also uses physical k here: `obs_model.cpp` computes `xx0 = x / (angpix * boxSize)` and hands it to `Zernike::Z_cart`, which doesn't renormalize. So `rlnOddZernike`/`rlnEvenZernike` are in rad·Å^n and can't currently be passed to torch-ctf as-is. cryoSPARC's basis is physical too.

Same beam tilt (0.5 mrad, Cs 2.7 mm, 300 kV) sampled at k = 0.2 cycles/Å on three grids, against the standard `2*pi*Cs*lambda**2*tilt*k**3`:

```
(128, 128) 1.5 A/px:   expected 0.263017  got 2.510746  ratio 9.546
(256, 256) 1.0 A/px:   expected 0.263017  got 0.743925  ratio 2.828
(384, 384) 0.65 A/px:  expected 0.263017  got 0.204300  ratio 0.777
```

The ratio is `1 / rho_max**3` every time, and it crosses 1.0, so you can't fold it into a constant. At 0.5 mrad the error is over a cycle of phase, so zeros actually move.

Fix is one line:

```diff
-    rho, theta = fftfreq_grid_polar(fft_freq_grid)
+    rho, theta = fftfreq_grid_polar(fft_freq_grid, normalize_rho=False)
```

`rho` only feeds `apply_odd_zernikes` and `apply_even_zernikes`, so nothing else depends on its scale, and `beam_tilt_to_zernike_coeffs` becomes correct as written. After the change all three grids give `0.263017`, matching RELION's `TiltHelper::insertTilt` too (opposite sign, since RELION stores the conjugate correction).

Do let me know if I've misunderstood, and users are meant to scale their coefficients into the normalized convention themselves. In that case the fix probably belongs in `beam_tilt_to_zernike_coeffs` instead then!

### DIFF
--- a/packages/primitives/torch-ctf/src/torch_ctf/_ctf_core.py
+++ b/packages/primitives/torch-ctf/src/torch_ctf/_ctf_core.py
@@ -163,7 +163,7 @@ def _build_freq_grid(
         device=device,
         transform_matrix=transform_matrix,
     )
-    rho, theta = fftfreq_grid_polar(fft_freq_grid)
+    rho, theta = fftfreq_grid_polar(fft_freq_grid, normalize_rho=False)
     return fft_freq_grid, fft_freq_grid_squared, rho, theta
 
 
--- a/packages/primitives/torch-ctf/tests/test_ctf_aberrations.py
+++ b/packages/primitives/torch-ctf/tests/test_ctf_aberrations.py
@@ -8,8 +8,10 @@ from torch_ctf import (
     apply_even_zernikes,
     apply_odd_zernikes,
     beam_tilt_to_zernike_coeffs,
+    calculate_relativistic_electron_wavelength,
     resolve_odd_zernikes,
 )
+from torch_ctf._ctf_core import _build_freq_grid
 
 
 def test_beam_tilt_to_zernike_coeffs():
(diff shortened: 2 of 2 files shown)


jdickerson95:
Great catch!

PR #112, opened 2026-08-12
feat: support Z42c/Z42s (secondary astigmatism) in apply_even_zernikes

joelyeois:
Following on from #111, still comparing cryoSPARC's aberration terms against torch-ctf.

I noticed `apply_even_zernikes` covers `Z44c`/`Z44s` and `Z60`, but there's nothing for the `n=4, m=±2` pair (secondary astigmatism) — passing it just raises `Unknown even Zernike`.

cryoSPARC gives you a four-component `ctf/tetra_A` per exposure group. Two of them are the `m=±4` terms and map onto `Z44c`/`Z44s` fine, but the other two are `m=±2` and there's nowhere to put them at the moment. I checked the whole set against cryoSPARC's simulator output and the basis is `k**4*cos(2θ)`, `k**4*sin(2θ)`, `k**4*cos(4θ)`, `k**4*sin(4θ)`, so the missing pair is just the `Z44` branch with `2*theta` instead of `4*theta`:

```diff
-        if name == "Z44c":  # 4-fold astigmatism
+        if name == "Z42c":  # secondary (2-fold) astigmatism
+            chi += coeff * rho**4 * torch.cos(2 * theta)
+        elif name == "Z42s":
+            chi += coeff * rho**4 * torch.sin(2 * theta)
+        elif name == "Z44c":  # 4-fold astigmatism
             chi += coeff * rho**4 * torch.cos(4 * theta)
```

Also added the supported names to the docstring, since I couldn't find them written down anywhere.

One test, following the existing `apply_even_zernikes` ones.

### DIFF
--- a/packages/primitives/torch-ctf/src/torch_ctf/ctf_aberrations.py
+++ b/packages/primitives/torch-ctf/src/torch_ctf/ctf_aberrations.py
@@ -308,8 +308,11 @@ def apply_even_zernikes(
     Parameters
     ----------
     even_zernikes: dict
-        Even Zernike coefficients. Values can be floats or tensors.
-        Floats will be converted to tensors for differentiability.
+        Even Zernike coefficients, keyed by name. Supported names are
+        ``Z42c``/``Z42s`` (secondary astigmatism), ``Z44c``/``Z44s``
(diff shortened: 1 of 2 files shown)


jdickerson95:
Looks great!

PR #113, opened 2026-08-12
fix(torch-fourier-slice) use edge subtraction before extraction

mgiammar:
@McHaillet @jdickerson95 

Updated code here to subtract the edge off the volume before padding (and use zero padding) rather than tracking the mean value and setting DC to zero in the FFT. The inward number of pixels/voxels the mean is averaged over is set to 4, but is an arbitrary choice.

Mean absolute errors improve significantly for some random Gaussian blobs where there is some non-zero background value added to the volume

```
C=  0.0  naive MAE from C*d =   0.0000   current MAE from C*d =   0.0000
C=  0.1  naive MAE from C*d =   0.8219   current MAE from C*d =   0.0000
C=  1.0  naive MAE from C*d =   8.2188   current MAE from C*d =   0.0000
C=  5.0  naive MAE from C*d =  41.0938   current MAE from C*d =   0.0000
C= 20.0  naive MAE from C*d = 164.3752   current MAE from C*d =   0.0000
C= 98.8  naive MAE from C*d = 811.6771   current MAE from C*d =   0.0005
```
image

### DIFF
--- a/packages/primitives/torch-fourier-slice/src/torch_fourier_slice/project.py
+++ b/packages/primitives/torch-fourier-slice/src/torch_fourier_slice/project.py
@@ -9,7 +9,7 @@ from .slice_extraction import (
     transform_slice_2d,
     transform_slice_2d_multichannel,
 )
-from .volume_utils import compute_cube_face_averages
+from .volume_utils import compute_cube_face_averages, compute_square_edge_averages
 
 
 def project_3d_to_2d(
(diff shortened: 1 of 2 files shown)


McHaillet:
Just for understanding: the problem is that having a boundary from the background value at the edges compared to the padded zeros will cause ringing, correct?

I like the fix; I am just trying to think if there are situations where this would be worse than doing nothing, but I don't think I can think of any. This works especially well if there is a protein in the center of the volume and the edges are clean. For a tomographic reconstruction that you project this is also fine because it won't make it worse than it is.

McHaillet:
BTW, the plot shows mean absolute error, what is the error when comparing the prediction and true image when normalising both?

jdickerson95:
I would also like to see some with and without DC 0 tests, in cases with high DC 0 , to make sure we don't get artefacts from that.

mgiammar:
[padding_mean_centering_comparison.pdf](https://github.com/user-attachments/files/31087217/padding_mean_centering_comparison.pdf)

@jdickerson95 @McHaillet see attached notebook, namely the plots at the end. This generates the "random Gaussian blobs" which admit an analytical expression for the ground-truth projection and compares different methods for Fourier slice extraction. The metric `L2` is relative L2 norm read as a percentage error, and metric `r` is the normalized cross-correlation between the images.

Using the edge subtraction method handles cases with large relative means on the volumes whereas all other methods have some sort of issue for non-zero mean values.

McHaillet:
Wow, thanks for the in-depth comparison! So, it seems setting the DC to 0 is actively bad :eyes:

And edge subtraction is the clear winner for any rotations.

McHaillet:
I also like the explicit zero padding for clarity.

McHaillet (approved):

McHaillet:
I was just thinking: it would be handy if this is controlled by a keyword argument that defaults to True. Because in some cases you might want to turn it off if you already applied a mask yourself that you know drops to zero.

mgiammar:
> if you already applied a mask yourself that you know drops to zero

This should subtract zero from the volume and the new lines effectively become a no-op, at least I think?

McHaillet:
I am just thinking off a mask that still is dropping to zero in the 4 edge pixels.

PR #116, opened 2026-08-18
Performance improvement to `torch-motion-correction`/`torch-fourier-filter` for large movies

mgiammar:
Updates across the frontend and backend of `torch-motion-correction` to improve runtime performance, decrease GPU memory overhead, and support CPU-resident storage of large super-resolution movies. Also includes memory efficiency updates to `torch-fourier-filter` updates for movie doseweighting again for large movies.

- Functions `correct_motion`, `correct_motion_fast`, and the `estimate_global_motion`/`estimate_motion_cross_correlation_patches` now accept a `device` parameter which controls compute device while movie tensor may be stored elsewhere.
- Movie patch extraction, normalization, masking, and Fourier filtering are now a one-time computation and cached in `estimate_motion_optimizer.py` rather than re-computed per optimization iteration.
- Added Fourier cropping during patch pre-computation which _massively_ reduces downstream compute costs during optimization (10-25x runtime improvement) without loss of information.
- Global/patch cross-correlation motion estimation now supports downsample factor for movies to reduce memory/compute requirement in place of sub-pixel accuracy.
- Pixel coordinate grid now computed once and re-used in `correct_motion`/`correct_motion_slow`.
- Added `dose_weight_normalization_grid` and `dose_weight_frame_chunk` to support chunked dose weighting of extremely large movies

### DIFF
--- a/packages/algorithms/torch-motion-correction/README.md
+++ b/packages/algorithms/torch-motion-correction/README.md
@@ -6,9 +6,13 @@
 [![CI](https://github.com/teamtomo/torch-motion-correction/actions/workflows/ci.yml/badge.svg)](https://github.com/teamtomo/torch-motion-correction/actions/workflows/ci.yml)
 [![codecov](https://codecov.io/gh/teamtomo/torch-motion-correction/branch/main/graph/badge.svg)](https://codecov.io/gh/teamtomo/torch-motion-correction)
 
-Motion estimation and correction in PyTorch.
+Movie motion estimation and correction using PyTorch. Implements a spline-based deformation field to model the beam-induced motion over movie collection.
 
-work in progress
+## Implemented algorithms
+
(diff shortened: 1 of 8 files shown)


jdickerson95 (commented):
Do we need to also remove Ln 233-234 in estimate_local_motion.py to support the device living on the CPU and chunks being sent to the CPU?
```
device = device if device is not None else image.device
image = image.to(device)
```
It's probably also time to take 'work in progress' out of the README

mgiammar:
> Do we need to also remove Ln 233-234 in estimate_local_motion.py to support the device living on the CPU and chunks being sent to the CPU?
> 
> ```
> device = device if device is not None else image.device
> image = image.to(device)
> ```
> 
> It's probably also time to take 'work in progress' out of the README

Good catch, both addressed in latest commit.

PR #118, opened 2026-08-24
perf(torch-image-interpolation): sample without broadcasting the image over sample points

### LINKED ISSUE #117: torch-image-interpolation: sample_image_2d/3d backward allocates O(n_points x image), making differentiable projection intractable
## Summary

`sample_image_2d` and `sample_image_3d` broadcast the **whole image** once per sample point before calling `F.grid_sample`. In the forward pass that is a free stride-0 view, so it looks harmless. In the **backward** pass, `grid_sample` must materialize a dense `(n_samples, c, *image_shape)` input-gradient before the broadcast reduces it away — so peak memory and runtime scale with `n_points x image_elems` instead of `n_points`.

This makes `torch_fourier_slice.project_3d_to_2d` effectively untrainable at realistic box sizes, since its backward goes through `sample_image_3d`.

## Location

```python
# torch_image_interpolation/image_interpolation_3d.py:64
image = einops.repeat(image, 'c d h w -> b c d h w', b=n_samples)

# torch_image_interpolation/image_interpolation_2d.py:65
image = einops.repeat(image, 'c h w -> b c h w', b=n_samples)
```

Both then sample a batch of 1x1x1 (or 1x1) grids. The comment above it explains the intent — "view our volume as a batch of n_samples multi-channel volumes ... this enables sampling arbitrarily shaped arrays of coords" — which works, but `grid_sample` already supports arbitrary point counts natively in the **grid's** spatial dims, without touching the batch dim.

## Measurements

Each row measured in a **separate process** (`ru_maxrss` is a process high-water mark, so measuring both variants in one process hides the difference). torch 2.6.0, CPU.

**3D** — volume `(64, 64, 33)` complex64 (a padded rfft volume as `torch-fourier-slice` produces), 4096 sample points:

| | peak RSS | time |
|---|---|---|
| `sample_image_3d` (current) | 4.20 GB | 1.207 s |
| batch=1, points in grid | **0.19 GB** | **0.001 s** |

**2D** — image `(512, 512)` float32, 4096 sample points:

| | peak RSS | time |
|---|---|---|
| `sample_image_2d` (current) | 4.28 GB | 1.800 s |
| batch=1, points in grid | **0.19 GB** | **0.012 s** |

A `torch.profiler` run over the 3D backward attributes it clearly:

```
aten::empty    17.02 Gb self mem   31 calls
aten::sum      60.9% self CPU       3 calls @ 1.25 s each
```

i.e. a large dense allocation followed by the reduction that collapses the broadcast.

## Downstream impact

For `project_3d_to_2d`, `n_points = B*K*N^2` and `image_elems ~ (2N)^3/2` (2x pad, rfft), so the backward is roughly `O(B*K*N^5)`. For a single 40-tilt cryo-ET particle:

| box N | current | with the fix |
|---|---|---|
| 32 | 40 GB | 2.3 MB |
| 64 | 1.2 TB | 17.2 MB |
| 128 | 40 TB | 133 MB |

On an M1 this OOM-kills somewhere between N=40 and N=48. The forward pass is fine throughout — it is specifically the backward.

## Proposed fix

Keep `batch=1` and put the sample points in the grid's spatial dimensions, which is what `grid_sample` is designed for:

```python
# instead of einops.repeat(image, 'c d h w -> b c d h w', b=n_samples)
image = image[None]                                    # (1, c, d, h, w)
coordinates = coordinates.reshape(1, 1, 1, n_samples, 3)
samples = F.grid_sample(
    input=image,
    grid=array_to_grid_sample(coordinates, array_shape=image.shape[-3:]),
    mode=interpolation,
    padding_mode='border',
    align_corners=True,
)                                                      # (1, c, 1, 1, n_samples)
samples = samples.reshape(samples.shape[1], n_samples).T   # (n_samples, c)
```

The existing complex handling (real/imag as channels) and the `inside` masking carry over unchanged; only the packing differs. The 2D case is the same with one fewer spatial dim.

I verified equivalence on both real and complex inputs:

- outputs: **exactly** equal (`max|d| = 0.00e+00`)
- gradients: equal to `max|d| = 5.96e-08` (float32) and `2.98e-08` (complex64)

A standalone reproducer (no downstream deps, runs as `python repro.py current|fixed|check`) is available if useful — happy to open a PR with the change and a regression test if you'd like it in that form.

## Environment

torch 2.6.0 | Python 3.13.15 | macOS 26.5.1 arm64 | teamtomo monorepo at 96c427f

shahpnmlab:
Fixes #117. @mgiammar you offered to look over a PR — here it is.

## The change

`sample_image_1d/2d/3d` broadcast the whole image to `(n, c, *image_shape)` before calling `grid_sample`, then sampled a batch of n 1-element grids:

```python
image = einops.repeat(image, 'c d h w -> b c d h w', b=n_samples)
coordinates = einops.rearrange(coordinates, 'b zyx -> b 1 1 1 zyx')
```

That broadcast is free in the forward pass — it is a stride-0 view. It is not free in the backward pass: `grid_sample` must materialise a dense `(n, c, *image_shape)` input gradient before the broadcast sums it away, so peak memory and runtime scale with `n * image_elems` instead of with `n`.

`grid_sample` already takes an arbitrary number of sample points in the **grid's** own spatial dimensions, which is where they belong. So the batch dimension stays at 1 and the points move into the grid:

```python
image = einops.rearrange(image, 'c d h w -> 1 c d h w')
coordinates = einops.rearrange(coordinates, 'b zyx -> 1 1 1 b zyx')
```

Only the packing changes. The complex-as-channels workaround, `padding_mode='border'`, `align_corners=True` and the out-of-range masking are all untouched, and the unpacking rearranges follow the new axis order.

Note this also covers **`sample_image_1d`**, which has the same broadcast at line 64 — I had only reported 2D and 3D in the issue.

## Effect

Sampling 4096 points from a `(64, 64, 33)` complex64 volume (a padded rfft volume as `torch-fourier-slice` produces), forward + backward, each row measured in a **dedicated process** since `ru_maxrss` is a process high-water mark. torch 2.6.0, CPU:

| | peak RSS | fwd + bwd |
|---|---|---|
| before | 2.53 GB | 3.704 s |
| after | **0.26 GB** | **0.003 s** |

For `torch_fourier_slice.project_3d_to_2d`, where `n = B*K*N²` and `image_elems ~ (2N)³/2`, this takes the backward from roughly `O(B*K*N⁵)` to `O(B*K*N²)`. On one 40-tilt cryo-ET particle that is ~40 GB → ~2.3 MB at N=32, and ~40 TB → ~133 MB at N=128.

## Equivalence

Checked patched vs. unpatched in separate processes across real, multichannel and complex inputs, in 1D/2D/3D, for every interpolation mode, with coordinates deliberately spilling outside the image so the masking path is covered:

- **outputs: exactly equal** (`max|Δ| = 0.00e+00`) on all 11 cases
- **gradients w.r.t. image and coordinates: exactly equal**, except `bicubic` 2D image-grad at `2.4e-07` — float32 accumulation order, ~1 ulp

The existing 42 tests pass unchanged.

## Tests added

`tests/test_backward_memory.py`:

- `test_image_is_not_broadcast_over_samples` — spies on `F.grid_sample` and asserts the input batch size is 1 regardless of `n_samples`. This pins the property directly rather than trying to assert on memory, so it is deterministic and instant. It fails on `main` (`assert [64] == [1]`).
- `test_gradients_flow_to_image_and_coordinates` — `gradcheck` in float64 over both arguments for all three dimensionalities, so the new packing cannot silently change the gradient.

---

@alisterburt — happy to look at the experimental Fourier slicing API separately; this one is deliberately confined to the existing Python API since, as you said, it is worth fixing while it is still there.

There is a second, independent blocker to backprop through `torch-fourier-slice` that this PR does **not** address — `extract_central_slices_rfft_3d` mutating the caller's `rotation_matrices` in place. I have opened that as a separate PR to keep the two reviewable on their own merits.

### DIFF
--- a/packages/primitives/torch-image-interpolation/src/torch_image_interpolation/image_interpolation_1d.py
+++ b/packages/primitives/torch-image-interpolation/src/torch_image_interpolation/image_interpolation_1d.py
@@ -45,7 +45,6 @@ def sample_image_1d(
     # set up for sampling with torch.nn.functional.grid_sample
     # coordinates: (..., ) -> (n, )
     coordinates, ps = einops.pack([coordinates], pattern='*')
-    n_samples = coordinates.shape[0]
     w = image.shape[-1]
 
     # handle complex input
@@ -58,15 +57,20 @@ def sample_image_1d(
 
     # For grid_sample to work with 1D, we need to add a dummy height dimension
     # and treat it as a 2D image with height=2
-    image = einops.repeat(image, 'c w -> c h w', h=2)
-
-    # Create batch of images, one per sample
-    image = einops.repeat(image, 'c h w -> b c h w', b=n_samples)
+    # torch.nn.functional.grid_sample is set up for sampling grids, so we sample
+    # a single grid of shape (1, n) containing all n coordinates.
+    #
+    # the sample points belong in the *grid's* spatial dimensions. broadcasting the
+    # image to (n, c, h, w) also produces correct results, but makes the backward
+    # pass materialise a dense (n, c, h, w) input-gradient before the broadcast
+    # sums it away - so peak memory and runtime scale with n * image_elems rather
+    # than with n. see https://github.com/teamtomo/teamtomo/issues/117
+    image = einops.repeat(image, 'c w -> 1 c h w', h=2)
 
     # Create grid coordinates
     # We need to add a dummy y coordinate (set to 0) to make it work with grid_sample
     dummy_y = torch.zeros_like(coordinates)
-    coords_2d = einops.rearrange([dummy_y, coordinates], "yx b -> b 1 1 yx")  # (b, 1, 1, 2)
+    coords_2d = einops.rearrange([dummy_y, coordinates], "yx b -> 1 1 b yx")  # (1, 1, b, 2)
 
     # Take the samples
     # We need to convert the coordinates to grid_sample format
@@ -87,10 +91,10 @@ def sample_image_1d(
 
     # Reconstruct complex valued samples if required
     if input_image_is_complex is True:
-        samples = einops.rearrange(samples, 'b (complex c) 1 1 -> b c complex', complex=2)
+        samples = einops.rearrange(samples, '1 (complex c) 1 b -> b c complex', complex=2)
         samples = utils.view_as_complex(samples.contiguous())  # (b, c)
     else:
-        samples = einops.rearrange(samples, 'b c 1 1 -> b c')
+        samples = einops.rearrange(samples, '1 c 1 b -> b c')
 
     # Set samples from outside of image to zero explicitly
     inside = torch.logical_and(coordinates >= 0, coordinates <= w - 1)
--- a/packages/primitives/torch-image-interpolation/src/torch_image_interpolation/image_interpolation_2d.py
+++ b/packages/primitives/torch-image-interpolation/src/torch_image_interpolation/image_interpolation_2d.py
@@ -47,7 +47,6 @@ def sample_image_2d(
     # set up for sampling with torch.nn.functional.grid_sample
     # shape (..., 2) -> (n, 2)
     coordinates, ps = einops.pack([coordinates], pattern='* yx')
-    n_samples = coordinates.shape[0]
     h, w = image.shape[-2:]
 
     # handle complex input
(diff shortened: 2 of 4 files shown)


alisterburt:
Thanks @shahpnmlab ! The experimental API is backed by kernels so won't be affected by this. Can you de-AI some of the comments? Regression tests likely also unnecessary but whatever :-)

joelyeois:
Was about to ask about this too since I also came across the same issue just a few days ago when trying to autograd through torch-fourier-slice! +1 for fixing this :)

PR #120, opened 2026-08-27
fix(torch-fourier-slice): reduce peak memory usage on sinc2 correction

mgiammar:
Uses views and broadcasting to avoid materializing a dense second volume for the sinc2 correction. Since the grid is separable along x/y/z, there's no issue viewing each correction along the dimensions.

Updated code since I ran into a new OOM error on GPU hardware after the sinc2 correction was added


jdickerson95 (approved):

McHaillet:
Great fix!

PR #121, opened 2026-08-28
Integrate package for electrostatic potential calculation into monorepo

mgiammar:
@volodymyr-masalitin @paloha I found some time this week to work on refactoring and migrating the `teamtomo/torch-calculate-electostatic-potential` package into the monorepo. Most of the underlying capabilities are the same, but I've updated the public-facing API somewhat.

Right, the two main functions are `calculate_scattering_potential_3d` and `calculate_scattering_potential_2d` which do the isotropic Gaussian approximations and summations. There are the additional classes `AtomStack` which controls the atom positions/identities/b-factors going in to the calculation and the `GridConfig` which is the local lattice for simulation.


jdickerson95:
Just looking through this and can work a little on the integration with other packages.
For this package specifically.

Could we clarify the units/normalization of the returned potential? 
The calculation appears to Fourier transform the Peng electron scattering factors directly, but it is not obvious to me that the result is yet in physical electrostatic-potential units (V).  My understanding is that this should be $(2\pi\hbar^2/(m_e e) \approx 47.88\ \mathrm{V.Å^2})$ when using Å as the length unit. It would also be helpful to explicitly document the units of both the 3D potential and the projected 2D potential.

mgiammar:
Correct that the Peng scattering factors have a coefficient which is neglected here, and that the units of length are Angstroms. Can add that as a constant and document units in each method.

Note the bond scaling factor (`BOND_SCALING_FACTOR = 1.043`) from ttsim3d is absent here.

jdickerson95:
I've integrated torch-structure-manipulation and have quite a few changes to integrate the simulation packages together.

Packages from @davidetorre99, @rsanchezgarc, @volodymyr-masalitin, @paloha, @joelyeois all have some changes.

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/utils/peng_model.py
```diff
@@ -1,42 +1,196 @@
...
+
+    parameters_a: Mapping[str, Sequence[float]]
+    parameters_b: Mapping[str, Sequence[float]]
+
+
+BondedScatteringFactorProviders = Mapping[str, BondedScatteringFactorTable]
+ScatteringFactors = (
+    Literal["peng_elemental", "peng_bonded"] | BondedScatteringFactorProviders
+)
+BondedFallback = Literal["elemental", "error"]
```
mgiammar, line 81: Having more than/less than 5 Gaussian fit parameters is possible. The `str` --> `Sequence` mapping is nice, but would be nice to have a method/attribute defining how many elements are in the sequence.

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/potential.py
```diff
@@ -69,13 +96,14 @@ def evaluate_gaussian_sum(
...
-        return numerator / (2**D * voxel_size.prod())
+        result = PENG_SCATTERING_TO_POTENTIAL * numerator / (2**D * voxel_size.prod())
     else:
         squared_distance = einops.reduce(diff**2, "... n k d -> ... n k 1", "sum")
         prefactor = a_expanded * (4 * torch.pi) ** (D / 2) * gw_expanded ** (D / 2)
         exponent = torch.exp(-4 * torch.pi**2 * squared_distance * gw_expanded)
+        result = PENG_SCATTERING_TO_POTENTIAL * (prefactor * exponent).sum(dim=-1)
 
-        return (prefactor * exponent).sum(dim=-1)
+    return cast("torch.Tensor", result)
```
mgiammar: Should not need cast if `mypy` types are being properly propagated

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/__init__.py
```diff
@@ -222,46 +245,54 @@ def fit_map_in_pdb(
...
-    simulator : DensitySimulator or None
-        Density simulator.  See :class:`~torch_fit_in_map.DensitySimulator`.
-        When ``None``, the default ``espcalculator``-based simulator is used.
+        Cubic box size for the simulated reference potential (voxels).
+    simulator : PotentialSimulator or None
+        Potential simulator.  See :class:`~torch_fit_in_map.PotentialSimulator`.
+        When ``None``, the default electrostatic-potential simulator is used.
+    simulator_config : PotentialSimulatorConfig or None
+        Options for the default simulator (scattering factors, sublattice radius,
+        etc.).  Ignored when a custom ``simulator`` is supplied.
```
mgiammar: Should have these options be either or so if two non-`None` values are supplied then a value error is raised.

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/__init__.py
```diff
@@ -279,64 +310,71 @@ def fit_map_in_pdb(
...
     box_size : int
-        Cubic box size for the simulated density (voxels).
-    simulator : DensitySimulator or None
-        Density simulator.  See :class:`~torch_fit_in_map.DensitySimulator`.
+        Cubic box size for the simulated potential (voxels).
+    simulator : PotentialSimulator or None
+        Potential simulator.  See :class:`~torch_fit_in_map.PotentialSimulator`.
+    simulator_config : PotentialSimulatorConfig or None
+        Options for the default simulator.  Ignored when a custom ``simulator``
+        is supplied.
```
mgiammar: Same as above, have options be either or

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/_atoms.py
```diff
@@ -56,7 +61,12 @@ def transform_atoms(
         A copy of *atoms* with the ``x``, ``y``, ``z`` columns transformed into
         the reference frame.
     """
-    import numpy as np
+    import torch
+    from torch_structure_manipulation import (
+        apply_rotation_to_coords,
+        apply_translation_to_coords,
+        df_to_atomxyz,
+    )
```
mgiammar: Move lazy imports to top of module

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/_gradient.py
```diff
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from typing import cast
+
```
mgiammar: Should not need a cast call if type hinting is being properly passed across functions

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/_simulate.py
```diff
@@ -64,99 +66,102 @@ def simulate(
...
-                "Install it with:\n"
-                "  pip install git+https://github.com/teamtomo/torch-calculate-electrostatic-potential.git\n"
-                "or pass a custom simulator= to fit_map_in_pdb / fit_pdb_in_map."
-            ) from exc
+        from torch_calculate_electrostatic_potential import (
+            GridConfig,
+            default_sublattice_radius,
+            potential_from_structure_3d,
+        )
+        from torch_structure_manipulation import AtomicStructure
```
mgiammar: Move imports to top of module

packages/algorithms/torch-fit-in-map/tests/test_dataframe_api.py
```diff
@@ -40,17 +40,17 @@ def _make_atoms() -> pd.DataFrame:
     )
 
 
-def test_fit_pdb_in_map_accepts_dataframe():
-    """fit_pdb_in_map should accept an atoms DataFrame and honor a custom simulator."""
-    from torch_fit_in_map import ExhaustiveSearchConfig, fit_pdb_in_map
+def test_fit_structure_in_map_accepts_dataframe():
+    """fit_structure_in_map accepts a DataFrame and honors a custom simulator."""
+    from torch_fit_in_map import ExhaustiveSearchConfig, fit_structure_in_map
```
mgiammar: Same lazy imports again, should fail at top of module if cannot import

packages/algorithms/torch-fit-in-map/tests/test_dataframe_api.py
```diff
@@ -66,17 +66,17 @@ def test_fit_pdb_in_map_accepts_dataframe():
     assert torch.allclose(result.rotation_matrix.cpu(), torch.eye(3), atol=0.2)
 
 
-def test_fit_map_in_pdb_accepts_dataframe():
-    """fit_map_in_pdb should accept a reference atoms DataFrame."""
-    from torch_fit_in_map import ExhaustiveSearchConfig, fit_map_in_pdb
+def test_fit_map_in_structure_accepts_dataframe():
+    """fit_map_in_structure accepts a reference-atoms DataFrame."""
+    from torch_fit_in_map import ExhaustiveSearchConfig, fit_map_in_structure
```
mgiammar: lazy imports

packages/algorithms/torch-fit-in-map/tests/test_dataframe_api.py
```diff
@@ -91,24 +91,23 @@ def test_fit_map_in_pdb_accepts_dataframe():
     assert isinstance(result.score, float)
 
 
-def test_transform_atoms_preserves_pairwise_distances():
-    """transform_atoms applies a rigid transform, preserving inter-atom distances."""
-    from torch_fit_in_map import AlignmentResult, transform_atoms
+def test_apply_alignment_to_structure_preserves_dataframe_and_distances():
+    """The structure transform preserves metadata and pairwise distances."""
+    from torch_fit_in_map import AlignmentResult, apply_alignment_to_structure
```
mgiammar: lazy imports

packages/algorithms/torch-fit-in-map/tests/test_dataframe_api.py
```diff
@@ -117,19 +116,113 @@ def _pdist(df):
...
     np.testing.assert_allclose(_pdist(atoms), _pdist(out), atol=1e-4)
+    assert list(out.columns) == list(atoms.columns)
+    assert out["label"].equals(atoms["label"])
 
 
-def test_transform_atoms_identity_centres_in_box():
+def test_apply_alignment_to_structure_identity_centres_in_box():
     """With identity rotation and zero shift, atoms are centred at the box centre."""
-    from torch_fit_in_map import AlignmentResult, transform_atoms
+    from torch_fit_in_map import AlignmentResult, apply_alignment_to_structure
```
mgiammar: lazy imports

packages/algorithms/torch-fit-in-map/tests/test_dataframe_api.py
```diff
@@ -117,19 +116,113 @@ def _pdist(df):
...
+    import torch_calculate_electrostatic_potential
+    from torch_calculate_electrostatic_potential import (
+        GridConfig,
+        potential_from_structure_3d,
+    )
+    from torch_structure_manipulation import AtomicStructure
+
+    from torch_fit_in_map import ExhaustiveSearchConfig, fit_structure_in_map
+    from torch_fit_in_map._simulate import DEFAULT_POTENTIAL_SIMULATOR
+    from torch_calculate_electrostatic_potential import default_sublattice_radius
```
mgiammar: lazy imports

packages/algorithms/torch-fit-in-map/tests/test_dataframe_api.py
```diff
@@ -117,19 +116,113 @@ def _pdist(df):
...
+
+    assert "torch_calculate_electrostatic_potential" in (
+        torch_calculate_electrostatic_potential.__file__ or ""
+    )
```
mgiammar: Why is this assertion statement needed? The name _should_ exist within the module and not something which should change

packages/algorithms/torch-fit-in-map/tests/test_dataframe_api.py
```diff
@@ -117,19 +116,113 @@ def _pdist(df):
...
+        gradient_config=None,
+        verbose=False,
+    )
+    assert np.isfinite(result.score)
+
+
+def test_default_simulator_supports_bonded_scattering_factors():
+    """Bonded Peng factors can be selected via PotentialSimulatorConfig."""
+    from torch_fit_in_map import PotentialSimulatorConfig
+    from torch_fit_in_map._simulate import DEFAULT_POTENTIAL_SIMULATOR
```
mgiammar: lazy imports

packages/algorithms/torch-fit-in-map/tests/test_exhaustive.py
```diff
@@ -117,16 +124,17 @@ def test_exhaustive_topk_returns_k_results():
 
 
 def test_fit_map_in_map_multistart():
-    """fit_map_in_map with n_start=3 should return the best NCC-scored refined result."""
+    """Multiple starts should return the best NCC-scored refined result."""
     from torch_fit_in_map import GradientRefinementConfig, fit_map_in_map
```
mgiammar: lazy imports

packages/algorithms/torch-fit-in-map/pyproject.toml
```diff
@@ -105,13 +109,30 @@ docstring-code-format = true
...
+    "torch_fourier_slice.*",
+    "torch_grid_utils",
+    "torch_grid_utils.*",
+    "torch_image_interpolation",
+    "torch_image_interpolation.*",
+    "torch_so3",
+    "torch_so3.*",
+]
+ignore_missing_imports = true
+
```
mgiammar, line 135: Verify how these mypy overrides interact with #130, if they are needed, or if they mask some other issues

packages/algorithms/torch-fit-in-map/README.md
```diff
@@ -11,16 +11,17 @@
...
 PDB/mmCIF files and the command-line tools live in the companion package
 [`torch-fit-in-map-cli`](https://github.com/rsanchezgarc/torch-fit-in-map-cli).
 
 Two alignment modes are supported:
 
-- **Map-to-map** — align one density map onto another (`fit_map_in_map`).
-- **Atoms ↔ map** — simulate an electrostatic-potential density from a table of
-  atoms and align it against a map (`fit_pdb_in_map` / `fit_map_in_pdb`).
+- **Map-to-map** — align one map onto another (`fit_map_in_map`).
+- **Atoms ↔ map** — simulate an electrostatic potential from a table of
```
mgiammar: Change custom arrow symbol to `<->`

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/utils/peng1996_element_params.npy
mgiammar, line 1: Should convert this to a `.json` file

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/utils/peng_model.py
mgiammar: re-visiting this after Joel's comment: not all parameterizations are Gaussian and therefore property should be renamed `n_terms`.

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/atom_stack.py
```diff
@@ -0,0 +1,193 @@
+"""Compatibility adapter from the historical AtomStack API to AtomicStructure."""
```
mgiammar, line 1: For now, I think this file is fine. But longer-term, we should decide which package "owns" the helper class for dealing with atomic structures. Currently there are `DataFrame` representations from `mmdf`, this `AtomStack`, the `AtomicStructure` class in `torch-structure-manipulation`, and the raw Tensor formats passed around in the backend.

Will create an issue so that we can follow up in the future.

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/potential.py
```diff
@@ -0,0 +1,326 @@
...
+    V_D(r) = C * sum_i a_i * (4*pi)^(D/2) * w_i^(D/2)
+             * exp(-4*pi^2 * w_i * |r|_D^2)
+
+The stored Peng coefficients already parameterize elastic electron scattering
+factors, not X-ray form factors:
+
+    f_e(s) = sum_i a_i * exp(-b_i * s^2),  s = sin(theta) / wavelength
+
+Consequently, they must not undergo the X-ray-to-electron Mott-Bethe conversion
+``f_e(s) = 0.023934 * (Z - f_X(s)) / s²`` again. To produce electrostatic
```
mgiammar: change to `s^2`

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/potential.py
```diff
@@ -0,0 +1,326 @@
...
+from typing import TYPE_CHECKING
+
+import einops
+import torch
+from tqdm import tqdm
+
+if TYPE_CHECKING:
+    from .grid import GridConfig
+
+# Electron-scattering-factor (Å) to Fourier-potential (V Å³) normalization.
```
mgiammar: Use `A` or `Angstroms` instead of special character, same with superscript

packages/primitives/torch-calculate-electrostatic-potential/README.md
```diff
@@ -0,0 +1,196 @@
...
+`torch_structure_manipulation.AtomicStructure`. The tensor-only
+`calculate_scattering_potential_2d` and `calculate_scattering_potential_3d`
+kernels remain public and support arbitrary leading batch dimensions.
+
+Coordinates and spacing are in Angstroms. Axis order is ZYX in 3D and YX in 2D.
+
+## Units and normalization
+
+`peng1996_element_params.npy` contains Peng et al. (1996) **elastic electron
+scattering factors**, not X-ray form factors:
```
mgiammar: Update README.md line upon conversion to JSON file

packages/primitives/torch-calculate-electrostatic-potential/README.md
```diff
@@ -0,0 +1,196 @@
...
+`f_e(s) = 0.023934 (Z - f_X(s)) / s²` is therefore already incorporated in the
+tabulated coefficients and must not be applied again.
+
+An electron scattering factor is not itself a real-space potential in volts.
+The package converts it to the Fourier transform of the electrostatic potential
+using
+
+```text
+V_tilde(g) = C f_e(g / 2),  g = 2s
+C = 2 pi hbar² / (m_e e) = 47.877647... V Angstrom²
```
mgiammar: Superscript notation, use s^2

packages/primitives/torch-structure-manipulation/src/torch_structure_manipulation/structure_transforms/rotate_molecule.py
```diff
@@ -0,0 +1,163 @@
...
+
+    return rotated_coords
+
+
+def create_rotation_matrix_from_euler(
+    angles: torch.Tensor,
+    order: str = "ZYZ",
+    degrees: bool = True,
+    device: torch.device | None = None,
+) -> torch.Tensor:
```
mgiammar: This function is just `roma.euler_to_rotmat` wrapped with some minimal checks. Would prefer to remove and instead call the `roma` function with proper device handling of tensors around it.

packages/primitives/torch-structure-manipulation/LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2020, TeamTomo
```
mgiammar, line 3: Copyright (c) 2020-2026, should apply this uniformly across all LICENSE files before next release

mgiammar:
Of note: compared to the JSON file with Peng scattering factors shipped in ttsim3d, these scattering factors in the JSON and numpy file differ for later elements: Ra, Ac, Th, Pa, U, Np, Pu, Am, Cm, Bk, Cf (Z=87–98).

Need to re-reference the Peng paper to see where/why these diverge. For all practical biological purposes, those atom scattering factors agree exactly.

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/atom_stack.py
mgiammar: Have made issue #133

mgiammar:
@jdickerson95 @McHaillet work here I think is largely done. Let me know if you want to check over anything. See my previous note on diverging scattering factors.

mgiammar:
> Of note: compared to the JSON file with Peng scattering factors shipped in ttsim3d, these scattering factors in the JSON and numpy file differ for later elements: Ra, Ac, Th, Pa, U, Np, Pu, Am, Cm, Bk, Cf (Z=87–98).
> 
> Need to re-reference the Peng paper to see where/why these diverge. For all practical biological purposes, those atom scattering factors agree exactly.

Looked over teh Peng 1996 paper and the erroneous scattering factors are in ttsim3d. Everything here matches, and ttsim3d should eventually migrate over to this package for scattering potential calculation.

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/utils/peng1996_element_params.json
```diff
@@ -0,0 +1,1392 @@
+{
+  "parameters_a": {
```
McHaillet, line 2: This file could be condensed by placing the lists on a single line instead of spreading. Similar to bonding .jsons

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/atom_stack.py
```diff
@@ -0,0 +1,193 @@
...
+    calculate_scattering_potential_3d,
+)
+from .utils.peng_model import get_peng_scattering_parameters
+
+if TYPE_CHECKING:
+    from .grid import GridConfig
+
+
+class AtomStack:
+    """Expose the historical atom-stack API over an ``AtomicStructure``."""
```
McHaillet, line 22: In what sense is this historical? It used to be called like this in ttsim3d or something?

McHaillet: Nvm, it's called like this in the torch-electrostatic-potentail repository...
Yes we should definitely consolidate one representation as you suggested in #133

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/grid.py
```diff
@@ -0,0 +1,436 @@
...
+    Dimensionality (2 or 3) is inferred from the length of `grid_shape`. Stencil is
+    computed and cached on first access.
+
+    Attributes
+    ----------
+    grid_shape : torch.Tensor
+        Shape of the grid in voxels, (ndim,) int64.
+    voxel_size : torch.Tensor
+        Size of each voxel in Angstroms, (ndim,) float32.
+    left_bottom_point : torch.Tensor
```
McHaillet, line 49: Why is the left bottom and right upper coordinate needed?

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/potential.py
```diff
@@ -0,0 +1,326 @@
...
+        half_voxel_expanded = einops.rearrange(half_voxel, "... d -> ... 1 1 d 1")
+        diff_expanded = einops.rearrange(diff, "... n k d -> ... n k d 1")
+
+        erf_diff = torch.erf(
+            (diff_expanded + half_voxel_expanded) * gamma_expanded
+        ) - torch.erf((diff_expanded - half_voxel_expanded) * gamma_expanded)
+        spatial_integral = torch.prod(erf_diff, dim=-2)
+        numerator = (spatial_integral * a_expanded).sum(dim=-1)
+
+        result: torch.Tensor = (
```
McHaillet, line 99: Why is the type hint here while it's not below in the else block?

packages/primitives/torch-calculate-electrostatic-potential/README.md
```diff
@@ -0,0 +1,196 @@
...
+
+```python
+from torch_calculate_electrostatic_potential import (
+    GridConfig,
+    potential_from_structure_2d,
+    potential_from_structure_3d,
+)
+from torch_structure_manipulation import AtomicStructure
+
+structure = AtomicStructure.from_dataframe(atoms, device="cuda")
```
McHaillet, line 71: What are the `atoms` here? What form does the dataframe need to have?

McHaillet (approved):
Awesome stuff, the code reads very very nicely

Just had a few questions, see the comments.

I also had a thought/question about the 3d sampling. It seems like you can specify different voxel size in x, y, and z. Does that mean you can also used the 3D variant to sample atoms directly to z-slices that you can use for multislice?

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/grid.py
mgiammar: Allows definition of the grid on non-centered coordinate systems. For example, if I wanted to rotate and re-simulate a structure I'd want that structure centered at `(0, 0, 0)`, then I could define the lower/upper bounds as `[-W/2,...]` and `[+W/2,...`. Otherwise, I'd need to translate and untranslated the 3D atomic coordinates if the lower left-hand corner is always `(0, 0, 0)`.

Mostly for convenience, currently not exercised in real code.

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/potential.py
mgiammar: There was previously a `cast("torch.Tensor", result)` statement here since somewhere above the typehinting lost the tensor attribute. This way mypy is still satisfied with the return type of the method.

IMO cast statement are harder to read in the code and indicative of some other type hinting issue. Had not spent time narrowing down which function/variable causes this lost tensor attribute here.

packages/primitives/torch-calculate-electrostatic-potential/README.md
mgiammar: `atoms` should be a dataframe read in from `mmdf`. Per the docstring:

> ```python
> """
> Required columns are ``x``, ``y``, ``z``, and ``element``. Atom names
>         come from ``atom`` when present. ``b_isotropic`` and ``occupancy``
>         default to zero and one, respectively. Optional ``bonded_environments``
>         and ``molecule_type`` columns are preserved when present.
> """
> ```

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/grid.py
McHaillet: Makes sense!

packages/primitives/torch-calculate-electrostatic-potential/src/torch_calculate_electrostatic_potential/potential.py
McHaillet: 100% agree with cast() being weird. That just feels like a bandaid to get mypy passing.

mgiammar:
> I also had a thought/question about the 3d sampling. It seems like you can specify different voxel size in x, y, and z. 
Does that mean you can also used the 3D variant to sample atoms directly to z-slices that you can use for multislice?

@McHaillet I think that should be possible somehow, though not tested. I've only tested isotropic pixel/voxel sizes as well as non-square/rectangular grids. But I bet there is a way to index stuff to sample the atoms to z-slices...

PR #122, opened 2026-09-01
fix(torch-ctf): support per-image Zernike coefficients

joelyeois:
Zernike coefficients don't work with batched inputs. Passing one coefficient
per image raises, where defocus, astigmatism, voltage, Cs, amplitude contrast
and phase shift all broadcast:

```python
calculate_ctf_2d(
    defocus=torch.linspace(1.0, 2.0, 3),
    astigmatism=torch.zeros(3),
    astigmatism_angle=torch.zeros(3),
    odd_zernike_coeffs={"Z33c": torch.full((3, 1, 1), 0.1)},  # per-image trefoil
    voltage=300.0, spherical_aberration=2.7, amplitude_contrast=0.1,
    phase_shift=0.0, pixel_size=1.0, image_shape=(64, 64),
    rfft=False, fftshift=False,
)
# RuntimeError: output with shape [64, 64] doesn't match the broadcast shape [3, 64, 64]
```

`apply_odd_zernikes` accumulates phase **in place** into a tensor sized from
the unbatched `(h, w)` grid. A coefficient of shape `(b, 1, 1)` makes the
right-hand side `(b, h, w)`, which an in-place add cannot expand into.

Real data hits this: trefoil and beam tilt are per-optics-group properties, and
a CryoSPARC `.cs` file stores them per particle.

`apply_even_zernikes` has the same defect, though it surfaces less often. It
starts from `total_phase_shift.clone()`, which is already batched whenever
defocus is, so the in-place add usually works. It still fails for a per-image
Zernike with a scalar defocus.

**Fix:** accumulate out of place in both functions. Scalar and already-batched
inputs behave as before, and only calls that used to raise now succeed. Tensor
coefficients already work for autograd (*allow tensor or float coefficients for
differentiability*, pre-monorepo); this extends them to batched shapes. The
float-to-tensor conversion duplicated in both loops moves into
`_as_zernike_coeff`, with the same `TypeError` message.

**Tests:** three added, all failing on `main` and passing here: per-image odd,
per-image even, and end to end through `calculate_ctf_2d`, following the
existing `test_2d_ctf_batch`. Full suite passes, ruff and mypy clean.

**Not changed:** a `(b,)`-shaped coefficient still fails, since Zernike
coefficients skip the `"... -> ... 1 1"` rearrange `_prepare_inputs` applies to
every other parameter. Adding it would change behaviour rather than fix a bug:
a `(1,)` coefficient would start returning `(1, h, w)`, which
`test_apply_odd_zernikes_with_beam_tilt` asserts against. A `(1,)`-shaped
*defocus* already returns `(1, h, w)` today. Happy to fold it in if you want
the consistency.

### DIFF
--- a/packages/primitives/torch-ctf/src/torch_ctf/ctf_aberrations.py
+++ b/packages/primitives/torch-ctf/src/torch_ctf/ctf_aberrations.py
@@ -227,6 +227,35 @@ def resolve_odd_zernikes(
     return zernikes
 
 
+def _as_zernike_coeff(
+    coeff: float | torch.Tensor, dtype: torch.dtype, device: torch.device
+) -> torch.Tensor:
+    """Return a Zernike coefficient as a tensor on the grid's dtype and device.
+
+    Parameters
+    ----------
+    coeff: float | torch.Tensor
+        Zernike coefficient. A scalar applies to every image; a tensor of
+        shape (..., 1, 1) gives one coefficient per image in a batch.
+    dtype: torch.dtype
+        Dtype of the coordinate grid the coefficient will be applied to.
+    device: torch.device
+        Device of the coordinate grid the coefficient will be applied to.
+
+    Returns
+    -------
+    coeff: torch.Tensor
+        The coefficient as a tensor.
+    """
+    if isinstance(coeff, int | float):
+        return torch.tensor(coeff, dtype=dtype, device=device)
+    if isinstance(coeff, torch.Tensor):
+        return coeff.to(device=device, dtype=dtype)
+    raise TypeError(
+        f"Zernike coefficient must be float or torch.Tensor, got {type(coeff)}"
+    )
+
+
 def apply_odd_zernikes(
     odd_zernikes: dict | None,
     rho: torch.Tensor,
@@ -240,8 +269,9 @@ def apply_odd_zernikes(
     Parameters
     ----------
     odd_zernikes: dict | None
-        Odd Zernike coefficients. Values can be floats or tensors.
-        Floats will be converted to tensors for differentiability.
+        Odd Zernike coefficients. Values can be floats or tensors. A
+        coefficient of shape (..., 1, 1) is treated as one value per image
+        and broadcasts the returned phase up to (..., h, w).
     rho: torch.Tensor
         Radial coordinate.
     theta: torch.Tensor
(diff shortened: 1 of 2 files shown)


packages/primitives/torch-ctf/src/torch_ctf/ctf_aberrations.py
```diff
@@ -227,6 +227,36 @@ def resolve_odd_zernikes(
...
+        Device of the coordinate grid the coefficient will be applied to.
+
+    Returns
+    -------
+    coeff: torch.Tensor
+        The coefficient as a tensor. Floats are converted with
+        ``requires_grad=True`` so that they remain differentiable.
+    """
+    if isinstance(coeff, int | float):
+        return torch.tensor(coeff, dtype=dtype, device=device, requires_grad=True)
```
mgiammar: Would all Zernike coefficients need gradient tracking? might be nice to have this as opt-in with another methods parameter.

Same idea for the `Tensor` path, if that tensor does not already require gradients, then never gains that attribute in this method.

packages/primitives/torch-ctf/src/torch_ctf/ctf_aberrations.py
```diff
@@ -268,29 +300,21 @@ def apply_odd_zernikes(
         return torch.zeros_like(rho)
 
     phase = torch.zeros_like(rho)
-    device = rho.device
-    dtype = rho.dtype
 
+    # Accumulation is out of place throughout: a per-image coefficient of
+    # shape (..., 1, 1) broadcasts the (h, w) accumulator up to (..., h, w),
+    # which an in-place `+=` into a (h, w) tensor cannot do.
```
mgiammar: Clean up comment, could be single-line about broadcast and out-of-place

packages/primitives/torch-ctf/src/torch_ctf/ctf_aberrations.py
```diff
@@ -326,31 +352,23 @@ def apply_even_zernikes(
         Phase shift with even Zernike coefficients applied.
     """
     chi = total_phase_shift.clone()
-    device = rho.device
-    dtype = rho.dtype
 
+    # Accumulation is out of place throughout: a per-image coefficient of
+    # shape (..., 1, 1) broadcasts an unbatched (h, w) total phase shift up to
+    # (..., h, w), which an in-place `+=` into a (h, w) tensor cannot do.
```
mgiammar: Same as above, condense comment

mgiammar (changes requested):

packages/primitives/torch-ctf/src/torch_ctf/ctf_aberrations.py
joelyeois: Actually it does look like there's no point in including the `requires_grad` parameter, because the user wouldn't really use it that way anyway. I just decided to remove it because it was redundant.

On the `Tensor` path — I left that as it was, since `Tensor.to` gives back the caller's own tensor when the dtype and device already match, so setting the flag there would mutate their input.

packages/primitives/torch-ctf/src/torch_ctf/ctf_aberrations.py
joelyeois: I removed the comments instead, since the other places also do not include comments about broadcasting, to be consistent.

PR #126, opened 2026-09-03
Feat/move tilt series loading and preprocessing to torch-tilt-series

McHaillet:
This PR moves tilt series loading and preprocessing from torch-tomogram to torch-tilt-series. Preprocessing has also been extended with a linear background ramp subtraction, and bandpass filter that by default removes the DC and everything beyond nyquist. 

Values for preprocessing can be tuned so that they are also more broadly useful for torch-tiltxcorr and torch-refine-tilt-axis-angle. Preprocessing for those packages, as well as for torch-reconstruct-tomogram, is a Boolean keyword so that already processed images can be passed.

@davidetorre99 please let me know what you think of these changes, especially if anything feels off to you compared to how you had it in mind!


packages/algorithms/torch-reconstruct-tomogram/src/torch_reconstruct_tomogram/reconstruct.py
```diff
@@ -113,7 +119,7 @@ def reconstruct_subvolume(
     points_zyx: torch.Tensor,
     sidelength: int,
     output_pixel_spacing: float | None = None,
-    normalize: bool = True,
+    preprocess: bool = True,
```
davidetorre99: What about adding some **kwargs here to change the preprocessing parameters on call? (same for reconstruct_tomogram and extract_particle_tilt_series)

davidetorre99:
It's great! I just added a minor comment to add some **preprocessing_kwargs to edit preprocessing parameters on call, but besides that, I really like it! Thanks!

McHaillet:
Added the **preprocessing_kwargs in!

PR #130, opened 2026-09-03
ci: enable pre-commit (ruff, mypy, typos, validate-pyproject) in CI

### LINKED ISSUE #79: CI tests do not include linting/formatting checks
Currently implemented CI pipeline only applies unit tests across the package, but does not include functionality for checking the code formatting standards.

Need to update the GitHub actions to include these checks for long-term maintainability and standardization.

McHaillet:
@mgiammar this is what a passing pre-commit looks like. We could still consider if we truly want to enforce all these rules. I would personally also be okay with just ruff and skip mypy/typos. 

---

Fixes #79.

I had Claude go through the codebase to make pre-commit passing. This means we can pre-commit to the CI and make sure PR's confirm to our formatting rules.

My previous PR #129 also enabled a check for the uv.lock. So hopefully these two things ensure some consistency.


mgiammar:
Thanks @McHaillet will take a look at this either tomorrow or next week.

I think having mypy/ruff skipped for now and slowly working new PRs to include the mypy/ruff checks could be a workable solution.

McHaillet:
I already let it go through the codebase to fix ruff mypy with out current pre-commit settings. So in principal we could enforce those style guides afterwards.

However, the main thing to think about is whether we want to fix it at these style guides, or if there is something we thing is overly stringent or whatever. This would be the time to change it.

jdickerson95:
Now that it's passing, I'm happy to keep mypy/typos.

mgiammar:
Ah didn't realize the PR had all of mypy/typos passing. I'd like to look over to make sure Claude didn't do any egregious re-formatting. But I agree with Josh if it's passing now we should enforce that as pre-commit going forward.

I misunderstood and thought mypy/typos wasn't passing yet.

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/_preprocess.py
```diff
@@ -73,7 +73,7 @@ def crop_or_pad_to_shape(
 
     # --- crop ---
     slices: list[slice] = [slice(None)] * (volume.ndim - 3)
-    for current, target in zip(volume.shape[-3:], target_shape):
+    for current, target in zip(volume.shape[-3:], target_shape, strict=False):
```
McHaillet: I am not sure why this was added... I think it should be strict=True because volume.shape[-3:] and target shape should be the same length, and zip should error if that's not the case.

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/_result.py
```diff
@@ -3,8 +3,10 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
+from typing import TYPE_CHECKING
 
-import torch
+if TYPE_CHECKING:
```
McHaillet, line 8: I haven't seen this before, but it seem like standard practice if you only need an expensive library for type checking.

packages/algorithms/torch-local-resolution/src/torch_local_resolution/estimate_local_resolution.py
```diff
@@ -208,7 +208,9 @@ def compute_correlation(inp: ComputeResolutionInput) -> torch.Tensor:
             low=low_cpp,
             high=high_cpp,
             falloff=inp.falloff,
-            image_shape=padded_image_shape,
+            image_shape=cast(
```
McHaillet: Also had not seen this before, it's legit for ensuring typing for linting, but it does not actually change the type.

packages/algorithms/torch-motion-correction/src/torch_motion_correction/estimate_motion_optimizer.py
```diff
@@ -92,7 +92,7 @@ def _prepare_patch_state(
     patch_shape = patch_sampling.patch_shape
     ph, pw = patch_shape
 
-    t, h, w = image.shape
+    _t, h, w = image.shape
```
McHaillet: _t is not used here, so perhaps this should be changed

```suggestion
    _, h, w = image.shape
```

packages/algorithms/torch-segment-fiducials-2d/src/torch_segment_fiducials_2d/dataset/augmentation.py
```diff
@@ -26,19 +26,21 @@ def random_crop_at_size(
...
-        images = [TF.vflip(image) for image in images]
-    return list(images)
+        result = [TF.vflip(image) for image in result]
+    return result
 
 
 def random_rotation(*images: torch.Tensor, p=0.5) -> List[torch.Tensor]:
     """Apply the same random rotation to multiple images."""
     angle = random.uniform(-180, 180)
+    result = list(images)
```
McHaillet, line 40: Slightly weird stuff going on here, but the explicit conversion to list makes sense given the rest of the code

packages/primitives/torch-fourier-rescale/src/torch_fourier_rescale/fourier_rescale_3d.py
```diff
@@ -78,29 +93,19 @@ def fourier_rescale_3d(
     dft = torch.fft.rfftn(image, dim=(-3, -2, -1))
     dft = torch.fft.fftshift(dft, dim=(-3, -2))
 
-    # Calculate target shape if using spacing
```
McHaillet, line 81: This is indeed redundant because it's already calculated above.

packages/primitives/torch-fourier-shift/src/torch_fourier_shift/dft_utils.py
```diff
@@ -39,16 +49,30 @@ def ifftshift_2d(input: torch.Tensor, rfft: bool) -> torch.Tensor:
 
 
 def fftshift_3d(input: torch.Tensor, rfft: bool) -> torch.Tensor:
+    """Apply fftshift to a 3D (r)fft-transformed tensor's last three dims."""
     if rfft is False:
         output = torch.fft.fftshift(input, dim=(-3, -2, -1))
     else:
-        output = torch.fft.fftshift(input, dim=(-3, -2,))
+        output = torch.fft.fftshift(
```
McHaillet: Not sure why this is formatted, should easily fit on one line.

packages/primitives/torch-grid-utils/src/torch_grid_utils/patch_grid/patch_grid.py
```diff
@@ -420,7 +420,7 @@ def _extract_patches_3d(self, key: int | slice | tuple | Any) -> torch.Tensor:
         # (gd, 1, 1, pd, 1, 1), (1, gh, 1, 1, ph, 1), etc.
 
         # Build the indexing arrays properly
-        gd, gh, gw = self.grid_shape
+        _gd, _gh, _gw = self.grid_shape
```
McHaillet: Maybe not used?

McHaillet (commented):
I've read all the way up to torch-image-interpolation. And highlighted things that seemed more severe than cosmetics.

McHaillet:
> I misunderstood and thought mypy/typos wasn't passing yet.

No worry!

Let's double check things and then we can merge.

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/_preprocess.py
McHaillet: Good catch, agreed - fixed to strict=True (both are always length 3, target_shape is typed tuple[int, int, int]), so a real mismatch now raises instead of silently truncating via zip's default behavior.

packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/_result.py
McHaillet: Yep, it's the standard pattern for deferring an import that's only needed for type hints (avoids paying its import cost / needing it as a runtime dependency at all when it's only used in annotations). Kept as-is.

packages/algorithms/torch-local-resolution/src/torch_local_resolution/estimate_local_resolution.py
McHaillet: Agreed on the general point (cast() doesn't add a runtime check) - went back through every cast() I'd added and replaced them with real narrowing: explicit length checks that raise for the two shape-arity cases where the value is genuinely variable (this one, and the equivalent in fsc.py/whitening.py), and for torch-fourier-rescale, unpacking into exactly N names + rebuilding a literal tuple (mypy infers the fixed-length type from that on its own, and it's a real runtime check - errors if the helper ever returns the wrong count). For this specific one: padded_image_shape's actual source (prepare_halfmaps_for_fft) was declared as bare tuple/tuple[int, ...] even though its own docstring + ndim validation guarantee 2 or 3 elements, so I tightened that return type instead and propagated it through, which made the cast here unnecessary entirely.

packages/algorithms/torch-motion-correction/src/torch_motion_correction/estimate_motion_optimizer.py
McHaillet: Fixed, thanks.

packages/algorithms/torch-segment-fiducials-2d/src/torch_segment_fiducials_2d/dataset/augmentation.py
McHaillet: Yeah, agreed it's a bit more verbose than ideal - happy to simplify further if you have a preferred shape for it, but this was the smallest change that kept *images as a tuple param (matching its declared type) while still being able to build up the transformed result across the two flip stages.

packages/primitives/torch-fourier-rescale/src/torch_fourier_rescale/fourier_rescale_3d.py
McHaillet: Confirmed and removed, thanks for double-checking.

packages/primitives/torch-fourier-shift/src/torch_fourier_shift/dft_utils.py
McHaillet: Removed the magic trailing comma in dim=(-3, -2,) - that's what was forcing ruff-format to keep it expanded. Collapses to one line now.

packages/primitives/torch-grid-utils/src/torch_grid_utils/patch_grid/patch_grid.py
McHaillet: Confirmed unused - removed the line entirely rather than just renaming.

packages/primitives/torch-subpixel-crop/src/torch_subpixel_crop/subpixel_crop_2d.py
```diff
@@ -68,74 +71,68 @@ def subpixel_crop_2d(
...
+    images: torch.Tensor,  # (batch, h, w)
+    positions: torch.Tensor,  # (n, batch, 2)
+    output_image_sidelength: int,
+    mask: Optional[torch.Tensor] = None,
+    return_rfft: bool = False,
+    decenter: bool = False,
 ) -> torch.Tensor:  # (n, batch, ph, pw)
     batch, h, w = images.shape
-    n_pos, batch_check, _ = positions.shape
+    _n_pos, batch_check, _ = positions.shape
```
McHaillet: _n_pos not used, could change to _

src/teamtomo/primitives/__init__.py
```diff
@@ -4,82 +4,82 @@
 try:
     import torch_affine_utils
 except ImportError:
-    torch_affine_utils = None
+    torch_affine_utils = None  # type: ignore[assignment]
```
McHaillet, line 7: Is this the way to handle it?

McHaillet (commented):
Went through the second part.

packages/primitives/torch-subpixel-crop/src/torch_subpixel_crop/subpixel_crop_2d.py
McHaillet: Fixed, thanks.

src/teamtomo/primitives/__init__.py
McHaillet: Good question - I actually went and tried the same real-narrowing approach as the cast() fixes (declare `X: ModuleType | None` before the try/import, so the except branch's `X = None` genuinely matches instead of needing an ignore). It works in isolation, but breaks once mypy analyzes the package as a whole: teamtomo/__init__.py does `from teamtomo import primitives, wip`, so when mypy follows that import graph (which is what actually happens here, not just checking the file standalone), it treats the bare annotation and the subsequent `import X` as two conflicting definitions of the same name (no-redef). So unlike the cast() cases, there isn't a clean real-narrowing alternative for this try/except-ImportError-fallback pattern - type: ignore[assignment] is the correct way to handle it here, and is also what mypy's own docs recommend for this exact optional-import idiom. Left it as-is.

McHaillet:
@mgiammar, I had a read through all of it. Let me know if you spot some more issues.

packages/algorithms/torch-fit-in-map/tests/test_exhaustive.py
```diff
@@ -12,10 +12,14 @@
 
 
 def test_align_identity_no_transform():
-    """Aligning a volume with itself should recover near-identity rotation and zero shift."""
+    """Align a volume with itself; should recover near-identity rotation and
+    zero shift."""
```
mgiammar: Docstring should be one-line title

packages/algorithms/torch-fit-in-map/tests/test_exhaustive.py
```diff
@@ -91,13 +95,18 @@ def test_symmetry_reduces_search_space():
...
+    ).shape[0]
+    n_c4 = get_uniform_euler_angles(
+        psi_step=step, theta_step=step, **r_c4._asdict()
+    ).shape[0]
     assert n_c4 < n_c1
 
 
 def test_exhaustive_search_with_symmetry():
-    """Search with C1 symmetry should recover near-identity rotation for self-alignment."""
+    """Search with C1 symmetry should recover near-identity rotation for
```
mgiammar: Same one-line comment as above

packages/algorithms/torch-fit-in-map/tests/test_exhaustive.py
```diff
@@ -117,16 +126,19 @@ def test_exhaustive_topk_returns_k_results():
 
 
 def test_fit_map_in_map_multistart():
-    """fit_map_in_map with n_start=3 should return the best NCC-scored refined result."""
+    """fit_map_in_map with n_start=3 should return the best NCC-scored refined
```
mgiammar: Same one-line comment as above

packages/primitives/torch-fourier-rescale/src/torch_fourier_rescale/fourier_rescale_3d.py
```diff
@@ -78,29 +91,20 @@ def fourier_rescale_3d(
...
     )
     new_shape = target_shape
 
     # transform back to real space and recenter
     dft = torch.fft.ifftshift(dft, dim=(-3, -2))
     if preserve_mean:
-        # we changed the number of elements in the FT so set norm='forward' to deactivate
+        # we changed the number of elements in the FT so set norm='forward' to
+        # deactivate
         # default fft normalization by 1/n and normalise by the correct factor
```
mgiammar, line 108: Reflow comments

PR #132, opened 2026-09-06
fix dense 3D sampling and add tilt-axis volume rotate

jdickerson95:
## Summary
- Avoid OOM in `torch-image-interpolation` when sampling dense coordinate grids (no per-point image repeat).
- Fix `torch-transform-image` rotate/shift matrix device mismatch on CUDA.
- Add `rotate_image_3d_about_tilt_axis` for cryo-ET-style tilts about an XY-plane axis.

## Test plan
- [x] `pytest packages/primitives/torch-image-interpolation/tests`
- [x] `pytest packages/primitives/torch-transform-image/tests`
- [x] Spot-check: `rotate_image_3d_about_tilt_axis(..., tilt_axis_angle=90)` matches `rotate_then_shift_image_3d(..., rotate_zyx=(0, θ, 0))`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/primitives/torch-image-interpolation/src/torch_image_interpolation/grid_sample_utils.py
+++ b/packages/primitives/torch-image-interpolation/src/torch_image_interpolation/grid_sample_utils.py
@@ -1,3 +1,5 @@
+"""Helpers converting array coordinates to ``grid_sample`` coordinates."""
+
 from typing import Sequence
 
 import torch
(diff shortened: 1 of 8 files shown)


McHaillet:
Hey @jdickerson95, I am slightly against further expanding this with finetuned function and rather rely on external packages to construct these wrappers. 

Do you think this is a more general useful function to add here?

McHaillet:
Okay, I now see it's needed for the image simulation. To me it makes more sense to put it in that package, because its quite niche

jdickerson95:
@McHaillet that makes sense. I had it there and moved it here. I agree it should probably live there.

Keep the first two fixes (OOM error and device mismatch) and remove the new function?

McHaillet:
Sounds good to me!

sjrothfuss:
Sounds good to me too!

McHaillet (approved):
LGTM!

mgiammar (approved):
Also LGTM

PR #134, opened 2026-09-11
fix(torch-fit-in-map): only pass config to the default simulator

rsanchezgarc:
Follow-up to #121.

## Problem
`fit_map_in_structure` and `fit_structure_in_map` always call `simulator.simulate(..., config=simulator_config)`, even when `simulator_config` is `None`. A custom simulator without a `config` parameter therefore fails with `TypeError: ... got an unexpected keyword argument 'config'`. That happens even though `simulator` and `simulator_config` are already mutually exclusive, so a custom simulator never needs `config`.

This breaks the slow test `test_structure_in_map_recovery`. CI doesn't run slow tests, which is why the failure went unnoticed.

## Changes
- **New helper:** `_simulate_structure`, a private function in `torch_fit_in_map/__init__.py`, used by both fit functions. It keeps the existing mutual-exclusion `ValueError` and passes `config` only to the default simulator.
- **Docstring:** the `PotentialSimulator` docstring now says custom simulators may omit `config`.
- **New fast tests in `tests/test_dataframe_api.py`:**
  - a custom simulator without `config` works with both fit functions
  - passing both `simulator` and `simulator_config` raises `ValueError`

## Testing
- **New test:** it fails on current `main` with the `TypeError` above and passes with this change.
- **Test suite:** `uv run pytest packages/algorithms/torch-fit-in-map` gives 35 passed.
- **mypy:** strict mypy is clean.
- **ruff:** `ruff check` and `ruff format --check` are clean on the changed files.
- **Slow test:** `test_structure_in_map_recovery` is re-running; I'll comment with the result.

https://claude.ai/code/session_01B8BYa1SnVKWtM4UqGSxns1

### DIFF
--- a/packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/__init__.py
+++ b/packages/algorithms/torch-fit-in-map/src/torch_fit_in_map/__init__.py
@@ -230,6 +230,38 @@ def apply_alignment(
     )
 
 
+def _simulate_structure(
+    atoms: pd.DataFrame,
+    simulator: PotentialSimulator | None,
+    simulator_config: PotentialSimulatorConfig | None,
+    pixel_size: float,
+    box_size: int,
+    device: torch.device,
+) -> torch.Tensor:
+    """Simulate *atoms* with a custom simulator or the configured default one.
+
+    ``config`` is only passed to the default simulator, so custom simulators
+    implementing ``simulate(atoms, pixel_size, box_size, device)`` without a
+    ``config`` parameter keep working.
+    """
+    if simulator is not None and simulator_config is not None:
+        raise ValueError(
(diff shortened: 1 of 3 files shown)


rsanchezgarc:
The slow test now passes on this branch (4853b87):

```
uv run pytest packages/algorithms/torch-fit-in-map/tests/test_integration.py -m "" -k test_structure_in_map_recovery
1 passed, 5 deselected in 1655.96s (0:27:35)
```

On current `main` it fails with `TypeError: _GaussianCaSimulator.simulate() got an unexpected keyword argument 'config'`.

mgiammar:
@rsanchezgarc good catch and thanks for the fix! The changes look good and I'll check the slow test passes before merging.

mgiammar (approved):

PR #137, opened 2026-09-11
feat(torch-reconstruct-tomogram): accept pre-loaded tilt images

jdickerson95:
## Summary

`reconstruct_subvolume()` and `reconstruct_tomogram()` both read their images from `tilt_series.image_path` before handing them to the private `_reconstruct_subvolume()`, which has always taken images directly. The capability was already there, only the public API hid it.

This adds an optional `images` argument to both, resolved through a shared `_resolve_images()` helper.

## Why

Reconstructing from image data that is *derived* rather than stored on disk currently means writing an intermediate MRC stack first. The motivating case is an even/odd split of dose-fractionated movies.

Noise2noise denoisers need two tomograms that share signal but carry independent noise. You get them by splitting each tilt's movie frames into even and odd sets, summing each, and reconstructing both halves **through one shared alignment**. Neither half stack is a file, and neither should have to become one just to be reconstructed.

Before:

```python
# both halves round-tripped through disk, and two TiltSeries kept in sync
mrcfile.write("even.mrc", even_images, overwrite=True)
mrcfile.write("odd.mrc", odd_images, overwrite=True)
even = reconstruct_tomogram(ts_even, shape, sidelength=256)
odd = reconstruct_tomogram(ts_odd, shape, sidelength=256)
```

After:

```python
# one movie per tilt -> two summed stacks, identically aligned
even_images = torch.stack([frames[0::2].sum(dim=0) for frames in corrected_movies])
odd_images = torch.stack([frames[1::2].sum(dim=0) for frames in corrected_movies])

# one TiltSeries, one alignment, two volumes
even = reconstruct_tomogram(tilt_series, shape, sidelength=256, images=even_images)
odd = reconstruct_tomogram(tilt_series, shape, sidelength=256, images=odd_images)
```

Reusing a single `TiltSeries` is the point. Aligning the halves independently would leave sub-pixel signal mismatch between them, which a denoiser trained on the pair responds to by blurring away exactly the detail you wanted to keep.

The same argument covers other derived inputs: dose-weighted or CTF-corrected stacks, simulated tilt series, or images already in memory from an upstream step.

## Semantics

- `images` replaces the **loading** step only. `preprocess` still applies, so several stacks passed through one `TiltSeries` are preprocessed identically. That is what the half-tomogram case requires.
- Supplied images must already be selected and ordered to match the geometry. `tilt_series.image_indices` is **not** applied to them, and a length mismatch against `tilt_angles` raises with a message saying so.
- The compute device is taken from `images.device`, so caller-supplied data stays where the caller put it.
- The new argument sits before `**preprocessing_kwargs` in both signatures, so existing positional and keyword call sites are unaffected. Nothing outside the package calls either function.

## Tests

Five added: equivalence against the disk path for both functions, the length-mismatch error, `preprocess` still applying to supplied images, and two stacks reconstructed through one shared alignment.

The preprocessing test also pins down that `preprocess_tilt_series_images()` does not mutate its input, so one loaded stack can safely be reused across several reconstructions.

## Checks

| Check | Result |
|---|---|
| `uv run pytest` | 1239 passed, 69 skipped |
| `uv run --group dev pre-commit run --all-files` | all hooks pass |
| `uv sync --locked` | clean, no dependency changes |


packages/algorithms/torch-reconstruct-tomogram/src/torch_reconstruct_tomogram/reconstruct.py
```diff
@@ -116,12 +116,46 @@ def _reconstruct_subvolume(
     return patches
 
 
+def _resolve_images(
+    tilt_series: TiltSeries,
+    images: torch.Tensor | None,
+    preprocess: bool,
+    preprocessing_kwargs: dict[str, Any],
```
McHaillet:
```suggestion
    **preprocessing_kwargs: Any,
```

packages/algorithms/torch-reconstruct-tomogram/src/torch_reconstruct_tomogram/reconstruct.py
```diff
@@ -141,14 +175,19 @@ def reconstruct_subvolume(
...
+      data is left wherever the caller put it
     - `**preprocessing_kwargs` are forwarded to `preprocess_tilt_series_images`,
       overriding any of its defaults (`low`, `high`, `falloff`,
       `bandpass_padding`, `subtract_background`, `normalize`) - see that
       function's docstring for details
     """
-    images = load_tilt_series_images(tilt_series)
-    if preprocess:
-        images = preprocess_tilt_series_images(images, **preprocessing_kwargs)
+    images = _resolve_images(tilt_series, images, preprocess, preprocessing_kwargs)
```
McHaillet:
```suggestion
    images = _resolve_images(tilt_series, images, preprocess, **preprocessing_kwargs)
```

packages/algorithms/torch-reconstruct-tomogram/src/torch_reconstruct_tomogram/reconstruct.py
```diff
@@ -203,14 +243,22 @@ def reconstruct_tomogram(
...
+      caller-supplied data is left wherever the caller put it
     - `**preprocessing_kwargs` are forwarded to `preprocess_tilt_series_images`,
       overriding any of its defaults (`low`, `high`, `falloff`,
       `bandpass_padding`, `subtract_background`, `normalize`) - see that
       function's docstring for details
     """
-    images = load_tilt_series_images(tilt_series)
-    if preprocess:
-        images = preprocess_tilt_series_images(images, **preprocessing_kwargs)
+    images = _resolve_images(tilt_series, images, preprocess, preprocessing_kwargs)
```
McHaillet:
```suggestion
    images = _resolve_images(tilt_series, images, preprocess, **preprocessing_kwargs)
```

McHaillet:
Few suggestions above to make kwargs passing consistent for readability.

nitpick: order the preprocess bool above the preprocess_kwargs. (i didn't do that correct to begin with)

Good PR, thanks for adding this in!

packages/algorithms/torch-reconstruct-tomogram/src/torch_reconstruct_tomogram/reconstruct.py
jdickerson95: Applied in 7e68085.

packages/algorithms/torch-reconstruct-tomogram/src/torch_reconstruct_tomogram/reconstruct.py
jdickerson95: Applied in 7e68085.

packages/algorithms/torch-reconstruct-tomogram/src/torch_reconstruct_tomogram/reconstruct.py
jdickerson95: Applied in 7e68085.

jdickerson95:
Thanks for the review. All three suggestions applied in 7e68085, plus the ordering nitpick.

`_resolve_images()` now takes `**preprocessing_kwargs: Any` and both call sites forward them the same way the public functions already do.

On the nitpick, I read it as wanting the two preprocessing arguments adjacent, so `preprocess` now sits directly above `**preprocessing_kwargs` in both public signatures, with the docstring bullets reordered to match. In `reconstruct_tomogram()` that moves it below `blend_margin`, and in `reconstruct_subvolume()` below the new `images`. Let me know if you meant something else and I'll redo it.

One thing worth flagging since it goes slightly beyond a docs change: that reorder shifts the positional slot of `preprocess` and `blend_margin`. Every call site in the repo passes them by keyword and nothing outside the package calls either function, so nothing breaks, but it is a public signature change.

Checks after the change:

| Check | Result |
|---|---|
| `uv run pytest` | 1239 passed, 69 skipped |
| `uv run --group dev pre-commit run --all-files` | all hooks pass |
| `uv sync --locked` | clean, no dependency changes |

McHaillet (approved):
Thanks for updating!

Issue #133, opened 2026-09-09
Standardize atomistic structure representations across package


mgiammar:
## General Issue

Currently, there are `pandas.DataFrame` representations of parsed PDB/mmCIF files from `mmdf`, the `AtomicStructure` class from package `torch-structure-manipulation`, and the new `AtomStack` class added in `torch-calculate-electrostatic-potential` in #121. Tensors for atom positions are also passed around on the backend which is fine. These structure representations are overlapping and different portions of the TeamTomo API touch each of them.

## Objectives

Determine what features a "atomic structure helper class" need to be useful across the API surface. And determine where that helper class should live (which package).

## Other thoughts

- Keep `mmdf` as dataframe only for user experience
- Add some helper class within the monorepo which can parse these datasframes to Tensor representations (tbd)

PR #142, opened 2026-09-14
chore(torch-fit-in-map): address #121 review follow-ups

rsanchezgarc:
Follow-ups from the #121 review of `torch-fit-in-map`. No behaviour changes beyond float32 rounding (see below).

## Changes
- **`cast()` removed:** the last `cast()` in `_exhaustive.py` is now an annotated assignment, matching how #121 handled the same pattern in `potential.py`. The underlying cause is that `torch-image-interpolation` has no `py.typed` marker, so mypy cannot see its return type; adding markers to the primitives is a separate change.
- **Imports moved to module level:** in `_preprocess.py`, `tests/test_integration.py` and `tests/test_preprocess.py`. `test_integration.py` imports `mrcfile` at module level, which until now was only present as a transitive dependency, so `mrcfile` is declared in this package's `test` group and `uv.lock` is updated.
- **mypy overrides documented:** the reviewer asked whether the `ignore_missing_imports` overrides are still needed. They are: none of the listed workspace modules ships a `py.typed` marker and `pandas-stubs` is not installed. Added a comment saying so rather than changing the list.
- **Duplicate centring removed:** `_geometry.center_positions_in_simulation_box` is gone; the simulator now uses `center_structure_from_coords` from `torch-structure-manipulation`, per the "which package owns structure helpers" discussion (#133).

## Note on the centring change
The two formulations differ only by float32 rounding: at most 1.9e-6 A in atom positions, and 1.9e-5 V in the simulated potential against a 10.7 V peak. That is enough to exceed the default `assert_close` tolerances, so `test_default_workspace_simulator_produces_zyx_volume_and_fits` now builds its expected structure with the same helper instead of the old inline formula. Tolerances were not loosened.

## Testing
- `uv run pytest packages/algorithms/torch-fit-in-map`: 35 passed.
- `pre-commit run --files <changed files>`: validate-pyproject, typos, ruff, ruff format and mypy all pass.
- `uv lock --locked`: no drift.
- `main` is merged into this branch (it includes #130 and #134); slow tests were not re-run here.
- The long line in `_config.py` was left alone; #130 already fixed it.

https://claude.ai/code/session_01B8BYa1SnVKWtM4UqGSxns1


McHaillet (approved):
Thanks for going over it and making it work with the new changes!

PR #143, opened 2026-09-16
more optimization for the fourier slice mojo backend

McHaillet:
Partly work to address issue #135

I have been continuing some work on the mojo experimental backed for torch-fourier-slice. One of the main things added here is a benchmark script that compares our canonical code (pytorch backend) to the new mojo backend, and optionally can also compare against torch-projectors (if its installed). 

### Benchmarking

The code runs a matrix that compares these things against each other:
* cpu/cuda
* slice insertion/slice extraction
* no_grad/backpropagation

Of note, backpropagation through the slice extraction path goes OOM for all the examples, precisely because of issue #118.

This is the full comparison, really cool that some of the CUDA paths now actually outspeed torch-projectors.

```
=== extraction / cpu / forward ===
  box    bp interp          canonical      experimental  torch_projectors
  128    41 cubic                  --         43.903 ms         25.266 ms
  128    41 linear          23.963 ms         16.614 ms         10.700 ms
  256    41 cubic                  --        160.162 ms        139.527 ms
  256    41 linear         127.748 ms         90.458 ms         82.935 ms
  256   200 cubic                  --        499.457 ms        397.766 ms
  256   200 linear         415.532 ms        124.014 ms        142.666 ms

=== extraction / cpu / forward_backward ===
  box    bp interp          canonical      experimental  torch_projectors
  128    41 cubic                  --        144.489 ms        110.484 ms
  128    41 linear                 --         53.689 ms         26.728 ms
  256    41 cubic                  --        564.195 ms        477.833 ms
  256    41 linear                 --        173.867 ms        155.714 ms
  256   200 cubic                  --       2177.019 ms       1938.713 ms
  256   200 linear                 --        456.930 ms        451.348 ms

=== extraction / cuda / forward ===
  box    bp interp          canonical      experimental  torch_projectors
  128    41 cubic                  --          2.931 ms          3.228 ms
  128    41 linear           7.328 ms          1.480 ms          2.478 ms
  256    41 cubic                  --          9.958 ms         14.682 ms
  256    41 linear          23.734 ms          2.675 ms          8.152 ms
  256   200 cubic                  --         46.744 ms        130.313 ms
  256   200 linear         109.108 ms          9.909 ms         17.416 ms

=== extraction / cuda / forward_backward ===
  box    bp interp          canonical      experimental  torch_projectors
  128    41 cubic                  --         17.532 ms         13.395 ms
  128    41 linear                 --          4.955 ms          3.728 ms
[...]

Issue #135, opened 2026-09-11
(torch-fourier-slice[mojo]): Add warp or block level local reductions for the slice insertion path.


mgiammar:
Tagging @alisterburt for visibility.

alisterburt:
Thanks @mgiammar 

This should be done for both slice and line insertion, all dimensionalities

PR #144, opened 2026-09-16
make mojo backend compatible with the mojo/max split

McHaillet:
The pin of mojo to v1.0.0 apparently broke our code. It was written against the beta v1.0.2b and not against the later stable release of v1.0.0. One that changed is that mojo moved gpu specific compilation to the Max package instead of pure mojo. We need to rely on the Max package to speed-up our code with warp and block level reductions (#135 and #143). 

This PR just ensures that compatibility by adding Max as a dependency and importing the correct stuff from there.

The other update that changed the most lines of code is that mojo 1.0.0 is more strict about type passing from cpu -> gpu. However, AFAIK the best way to do it is with DeviceParams, so we know use those to move kernel parameters around (things like box size, oversampling factors, boolean toggles).


McHaillet:
@alisterburt, would appreciate your thoughts on this, you might know something more about the design

McHaillet:
The compatibility for the mojo/max split is fixed. I wasn't entirely sure about the DeviceParams back and forth conversion, but apparently this is really needed to have fixed types for the move from host -> device. Afterward we want Int again (instead of Int32/64) because it can use the device native ideal integer size for indexing (same idea as size_t in C).

alisterburt:
Could you write this one up as an issue in modular/modular? Sounds worth raising and I don't know enough about that machinery

PR #148, opened 2026-09-19
fix(torch-fourier-slice): pose-gradient kernels returned zero gradients on Apple GPUs

alisterburt:
On Apple Silicon (mps) every `directions` / `rotations` / `shifts` gradient from the experimental Mojo kernels was exactly zero; CPU was correct.

**Root cause (Mojo compiler, Metal target):** by-reference variadic arguments lose their address space when spilled. The `*_pose_grad` pixel ops built the incoming cotangent as `C2(ptr[off], ptr[off + 1])`; `SIMD.__init__` takes its lanes by reference, the pack is spilled to an `alloca [2 x ptr]`, and on AIR the reloaded flat pointers into the device buffer are dereferenced as private memory, yielding zeros. Upstream issue: modular/modular#____ (Mojo-only reproducer + device IR; reproduces on 1.0.0 and nightly 1.1.0).

**Fix:** read the cotangent through the same `TileTensor` views the forward and scatter kernels use (the value is loaded before anything is passed by reference), in all six sites: forward/backproject × 3D line, slice, 2D line.

this is a workaround for modular/modular#7158

Verified: `test_gpu_pose_weight_gradients_match_cpu`, `test_gpu_line_pose_weight_gradients_match_cpu`, `test_gpu_2d_line_grads_match_cpu` pass on an M4 Pro; MPS-vs-CPU direction/rotation gradients agree to 1e-7.


alisterburt:
Root-cause data (sentinels inside `_forward_line_pose_grad_pixel`, M4 Pro, mojo 1.0.0 / max 26.5):

- `C2(grad_line[off], grad_line[off + 1])` -> reads **zeros** (original).
- two separate scalar loads into locals, then `C2(a, b)` -> reads **garbage** (MPS/CPU rel. err 1.4).
- explicit `grad_line.load[width=2](off)` -> same garbage.
- `TileTensor(grad_line + base, row_major(bp, w, 2))[i_bp, x, c]` -> correct (this PR).
- a lone sentinel `fetch_add(out, grad_line[off])` for one thread -> correct value.

So it is not the address, residency or the atomics: any raw `UnsafePointer` read of this buffer in the hot path returns stale data on Metal while `layout` loads of the same address are right; raw reads of the `direction` buffer in `_line_k` are fine. Looks like a Mojo Metal backend issue in how raw pointer loads are lowered (address space / caching) — worth a minimal repro upstream; tile loads are the safe convention meanwhile.

alisterburt:
Pinned it with a Mojo-only reproducer (no torch): inside a Metal kernel, a SIMD built **directly from inline pointer subscripts** reads zeros —

```mojo
var gp = C2(src[off], src[off + 1])        # (0, 0) on Metal
var a = src[off]; var b = src[off + 1]      # correct
var gp = C2(a, b)
var gp = src.load[width=2](off)             # correct
```

CPU is right in every form. Mojo 1.0.0 / max 26.5.0 / M4 Pro. That is exactly the form all six `*_pose_grad` cotangent reads used. Filing upstream on modular/modular with the standalone script; this PR's tile-view loads remain the right fix on our side.

PR #149, opened 2026-09-19
fix(torch-fourier-slice): declare torch-ctf dependency

alisterburt:
`torch_fourier_slice._grids._ewald_curvature` imports `torch_ctf` (for the electron wavelength), but `torch-fourier-slice` does not declare it, so installing the package outside the workspace fails with `ModuleNotFoundError: torch_ctf` on import. Adds it to `dependencies`. The lockfile may need regenerating (`uv lock`).


alisterburt:
cc @mgiammar should I specify versions here or nah?

mgiammar:
No, not declaring versions is fine. Removing the above `>=` version requirements in `pyproject.toml` would be good too

PR #150, opened 2026-09-19
fix(torch-fourier-slice): make extract/insert exact adjoints at Nyquist

alisterburt:
Fixes a minor difference in edge handling between scatter/gather operations which made the slice extraction/insertion paths not exact adjoints at Nyquist

note that the behavior after this is zero padding. @mgiammar probably edge padding or something smarter is better...

LLM slop below:
The forward interpolation stencils clamped out-of-range voxels while the scatter drops them, so `extract` / `insert` were not exact adjoints wherever a stencil reaches past the grid (the Nyquist shell): `<extract(V), y>` vs `<V, insert(y)>` differed by ~3% at the default cutoff (exact below it). The insertion's gradient gather already zero-padded via a separate `_sample_rfft_*_drop` sampler selected by a `drop` flag.

This makes zero-padding the one convention (it is also the canonical layer's): the zero-padding samplers become `_sample_rfft_3d` / `_sample_rfft_2d`, the clamp variants and the `drop` plumbing in the gradient gathers are removed, and the README layout note is updated. The adjoint identity now holds at full Nyquist; all other experimental tests unchanged (the three `test_gpu_*grad*` failures on Apple GPUs are #148).


McHaillet:
Thanks for fixing this :+1: 

Just wanted to check this:

> note that the behavior after this is zero padding. @mgiammar probably edge padding or something smarter is better...

But it is zero padding in Fourier space, correct?

alisterburt:
Yeah that's correct, which dampens gathered/scattered components at nyquist along x and y, nbd I think but worth noting

McHaillet:
thanks for confirming :)

